PR #30, opened 2024-05-14
added first pass of DP bundle and helpers

bhusang:
Will add example code although it should be simple...

### DIFF
--- a/examples/protocols/displayport/displayport-board.stanza
+++ b/examples/protocols/displayport/displayport-board.stanza
@@ -1,0 +1,202 @@
+#use-added-syntax(jitx)
+defpackage jsl/examples/protocols/displayport/displayport-board :
+  import core
+  import jitx
+  import jitx/commands
+
+  import jsl/protocols/displayport
+
+
+public pcb-rules displayport-board-rules :
+  min-copper-width             = 0.025
+  min-copper-copper-space      = 0.025
+  min-copper-hole-space        = 0.025
+  min-copper-edge-space        = 0.254
+  min-annular-ring             = 0.050
+  min-drill-diameter           = 0.100
+  min-silkscreen-width         = 0.100
+  min-pitch-leaded             = 0.350
+  min-pitch-bga                = 0.350
+  max-board-width              = 406.4
+  max-board-height             = 558.8
+  min-silk-solder-mask-space   = 0.050
+  min-silkscreen-text-height   = 0.380
+  solder-mask-registration     = 0.050
+  min-th-pad-expand-outer      = 0.050
+  min-soldermask-opening       = 0.152
+  min-soldermask-bridge        = 0.102
+  min-hole-to-hole             = 0.100
+  min-pth-pin-solder-clearance = 0.500
+
+;==== Materials ================================================================
+pcb-material copper :
+  type = Conductor
+  name = "Copper"
+
+pcb-material core :
+  type = Dielectric
+  dielectric-coefficient = 3.6
+  name = "FR4 Core"
+
+pcb-material soldermask :
+  type = Dielectric
+  dielectric-coefficient = 3.7
+  name = "Taiyo BSN4000"
+
+pcb-material prepreg :
+  type = Dielectric
+  dielectric-coefficient = 3.5
+  name = "FR4 Prepreg"
+
+
+public pcb-stackup displayport-stackup :
+  name = "Example 6-layer 0.032in"
+  layer(0.013, soldermask)
+  layer(0.030, copper    )
+  layer(0.060, prepreg   )
+  layer(0.025, copper    )
+  layer(0.060, prepreg   )
+  layer(0.015, copper    )
+  layer(0.070, core      )
+  layer(0.025, copper    )
+  layer(0.060, prepreg   )
+  layer(0.025, copper    )
+  layer(0.060, prepreg   )
+  layer(0.030, copper    )
+  layer(0.013, soldermask)
+
+public pcb-via gen-via (via-start:LayerIndex via-stop:LayerIndex via-name:String) :
+  name = via-name
+  start = via-start
+  stop = via-stop
+  diameter = 0.200
+  hole-diameter = 0.100
+  type = LaserDrill
+  via-in-pad = true
+
+; public pcb-routing-structure single-ended :
+;   name = "50.0 Ohm single-ended"
+;   layer(Top) :
+;     trace-width = 0.050     ; mm
+;     clearance = 0.100       ; mm
+;     velocity = 0.19e12      ; mm/s
+;     insertion-loss = 0.008  ; db/mm @ 1GHz
+;     neck-down = NeckDown(0.075 0.075 false false)
+
+;   layer(LayerIndex(2)) :
+;     trace-width = 0.050     ; mm
+;     clearance = 0.100        ; mm
+;     velocity = 0.19e12      ; mm/s
+;     insertion-loss = 0.008  ; db/mm @ 1GHz
+;     neck-down = NeckDown(0.075 0.075 false false)
+
+;   layer(LayerIndex(4)) :
+;     trace-width = 0.050     ; mm
+;     clearance = 0.100       ; mm
+;     velocity = 0.19e12      ; mm/s
+;     insertion-loss = 0.008  ; db/mm @ 1GHz
+;     neck-down = NeckDown(0.075 0.075 false false)
+
+; public pcb-differential-routing-structure differential :
+;   name = "100 Ohm common-mode impedance"
+;   layer(Top) :
+;     trace-width = 0.100     ; mm
+;     pair-spacing = 0.075    ; mm
+;     clearance = 0.100       ; mm
+;     velocity = 0.19e12      ; mm/s
+;     insertion-loss = 0.008  ; db/mm @ 1GHz
+;     neck-down = DifferentialNeckDown(0.080 0.080 0.100 false false)
+
+;   layer(LayerIndex(2)) :
+;     trace-width = 0.110     ; mm
+;     pair-spacing = 0.080    ; mm
+;     clearance = 0.120       ; mm
+;     velocity = 0.19e12      ; mm/s
+;     insertion-loss = 0.008  ; db/mm @ 1GHz
+;     neck-down = DifferentialNeckDown(0.080 0.080 0.100 false false)
+
+;   layer(LayerIndex(4)) :
+;     trace-width = 0.100     ; mm
+;     pair-spacing = 0.075    ; mm
+;     clearance = 0.100       ; mm
+;     velocity = 0.19e12      ; mm/s
+;     insertion-loss = 0.008  ; db/mm @ 1GHz
+;     neck-down = DifferentialNeckDown(0.080 0.080 0.100 false false)
+
+;   uncoupled-region = single-ended
+
+; Define the shape/size of the board
+public val board-shape = RoundedRectangle(50.0, 30.0, 0.25)
+
+; Creates a pcb-board with default stackup given the number of layers and board outline.
+public pcb-board displayport-example-board (outline:Shape) :
+  stackup = displayport-stackup
+  boundary = outline
+  signal-boundary = outline
+  vias = [gen-via(LayerIndex(0,Top), LayerIndex(1,Top) "uTop-1") ; Gnd
+          gen-via(LayerIndex(0,Top), LayerIndex(2,Top) "uTop-2") ; Signal
+          gen-via(LayerIndex(0,Top), LayerIndex(3,Top) "uTop-3") ; Gnd
+          gen-via(LayerIndex(0,Top), LayerIndex(4,Top) "uTop-4") ; Signal
+          gen-via(LayerIndex(0,Top), LayerIndex(0,Bottom) "th") ; Signal
+        ]
+
+; Setup the board
+public defn setup-board () :
+  set-board(displayport-example-board(board-shape))
+  set-rules(displayport-board-rules)
+
+doc: \<DOC>
+The pcb-differential-routing-structure template below is an example of a differential-routing-structure needed
+by pcie. This pcb-differential-routing-structure will be applied to the differential traces
+in the PCIe bus by using the helper function `pcie-apply-routing-structure()`.
+It should also reference a single ended routing structure for the uncoupled region.
+<DOC>
+public pcb-differential-routing-structure diff :
+  name = "100 Ohm differential impedance - PCIe"
+  val lookup-trace-impedance = displayport-get-trace-impedance()
+  val tw = switch(typ(lookup-trace-impedance)) :
+    85.0  : 0.1400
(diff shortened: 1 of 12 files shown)


examples/protocols/displayport/displayport-board.stanza
```diff
@@ -0,0 +1,202 @@
+#use-added-syntax(jitx)
+defpackage jsl/examples/protocols/displayport/displayport-board :
+  import core
+  import jitx
+  import jitx/commands
+
+  import jsl/protocols/displayport
+
+
+public pcb-rules displayport-board-rules :
```
callendorph, line 10: Is there a way we can reuse the board setup from the PCIe example so that we don't have this copy-pasta ? 

I create the PCIe board because the board that I had in `examples/landpatterns/board.stanza` was not sufficient for the PCIe signal integrity setup. 

But it seems like PCIe and Display-Port could share the stackup at least and then maybe have slightly different routing structures ?

examples/protocols/displayport/displayport-main.stanza
```diff
@@ -0,0 +1,116 @@
...
+  import jsl/landpatterns/two-pin/SMT
+
+  import jsl/protocols/displayport
+  import jsl/si/helpers
+
+  import jsl/examples/protocols/displayport/displayport-board
+  import jsl/examples/protocols/displayport/displayport-src
+
+
+public pcb-component block-cap :
```
callendorph, line 23: This is duplicated from PCIe - lets setup a common file that can be referenced from all si protocol examples. 

We might also need to add the `value` parameter - I don't see any for this component.

We also might consider making it parametric so that we don't have to create multiple copies for different values used by different application.

examples/protocols/displayport/displayport-board.stanza
bhusang: I suppose that we can do that -- my goal is to have a standalone example when users want to adapt the code to use it in their designs. If we move the board code out of each example, then they will have to root around to find it to copy it.

examples/protocols/displayport/displayport-src.stanza
```diff
@@ -0,0 +1,143 @@
...
+<DOC>
+
+val body = PackageBody(
+  width = 15.0 +/- 0.1,
+  length = 15.0 +/- 0.1,
+  height = 0.71 +/- [0.06, 0.0]
+)
+
+val BGA-pkg = BGA(
+  num-leads = 49,
```
callendorph, line 31: Leads is wrong here - should be 64. Probably wrong in PCIe as well.

examples/protocols/displayport/displayport-board.stanza
bhusang: And I suspect that for the first examples that we are going to have a lot of code-sharing, but I can see later examples might diverge especially if we get via-structures.

examples/protocols/displayport/displayport-src.stanza
bhusang: Will fix.

examples/protocols/displayport/displayport-board.stanza
callendorph: I think that kind of duplication is fine `jitx-cookbook`

I don't want to have to maintain duplication in JSL.

examples/protocols/displayport/displayport-src.stanza
```diff
@@ -0,0 +1,143 @@
...
+  val ctl-r = Ref("aux")
+  val pwr-r = Ref("pwr")
+
+  val box-params = BoxSymbolParams(
+    show-grid = false
+    col-width = 1
+  )
+
+  val box = BoxSymbol(self, params = box-params)
+  set-grid(box, [2, 2], ctl-r)
```
callendorph, line 90: You set a grid here but then you don't set any rows. If you aren't using the rows/columns then this grid is unnecessary.

examples/protocols/displayport/displayport-src.stanza
```diff
@@ -0,0 +1,143 @@
...
+    for i in 0 to 4 do :
+      if source :
+        displayport.lane[i].P => sw.DP-TX-P[i]
+        displayport.lane[i].N => sw.DP-TX-N[i]
+      else :
+        displayport.lane[i].P => sw.DP-RX-P[i]
+        displayport.lane[i].N => sw.DP-RX-N[i]
+    displayport.aux.P => sw.AUX-P
+    displayport.aux.N => sw.AUX-N
+    displayport.hpd   => sw.HPD
```
callendorph, line 143: Based on `source` - this means that either `sw.DP-RX-P/N` or `sw.DP-TX-P/N` will be connected but not both. Is that right ?

examples/protocols/displayport/displayport-src.stanza
bhusang: Gotcha. Removed.

examples/protocols/displayport/displayport-src.stanza
bhusang: Yes, DisplayPort is a unidirectional protocol. So I created a component that can be either a source or sink but not both.

examples/protocols/displayport/displayport-main.stanza
bhusang: Ok, I will try to create a common directory and then move the board and selected components to that common directory.

examples/protocols/displayport/displayport-src.stanza
```diff
@@ -0,0 +1,143 @@
...
+    [AUX-N      | C[5] ]
+    [AUX-P      | D[5] ]
+    [HPD        | D[6] ]
+
+    [VDD        | E[1] ]
+    [GND        | E[2] ]
+
+  val bus-r = Ref("bus")
+  val ctl-r = Ref("aux")
+  val pwr-r = Ref("pwr")
```
callendorph, line 82: I learned a neat trick yesterday. It isn't documented so probably shouldn't include it yet. 

`#R(bus)` will map to `Ref("bus")`. 

Provides a short hand for creating refs - especially indexed and field refs.

src/protocols/displayport.stanza
```diff
@@ -0,0 +1,220 @@
...
+  DP1p4  ; 32.40 Gbit/s / 4 lanes =>  8.1 Gbit/s / 10 UI =>  810 MHz clk
+  DP1p4a ; 32.40 Gbit/s / 4 lanes =>  8.1 Gbit/s / 10 UI =>  810 MHz clk
+  DP2p0  ; 80.00 Gbit/s / 4 lanes => 20.0 Gbit/s / 10 UI => 2000 MHz clk
+  DP2p1  ; 80.00 Gbit/s / 4 lanes => 20.0 Gbit/s / 10 UI => 2000 MHz clk
+  DP2p1a ; 80.00 Gbit/s / 4 lanes => 20.0 Gbit/s / 10 UI => 2000 MHz clk
+
+doc: \<DOC>
+@brief DisplayPort Bundle
+The Data Bundle consists of 4 lane pairs (1 RX or TX pair set = 1 lane)
+The bundle also constains an AUX channel (differential pair) and a HPD pin
```
callendorph, line 56: spelling:

`constains` => `constrains`

callendorph: edit - actully should be `contains` I guess

src/protocols/displayport.stanza
```diff
@@ -0,0 +1,220 @@
...
+
+public pcb-bundle displayport :
+  name = "DisplayPort"
+  description = "DisplayPort Digital Display Interface"
+  port aux  : diff-pair
+  port lane : diff-pair[4]
+  port hpd  : pin ; hot plug detect
+
+
+defn constrain-ch (sw:Toleranced, ml:Double, a:JITXObject, b:JITXObject) :
```
callendorph, line 67: Duplicate - lets create a function in `si/helpers.stanza` that implements this concept.

bhusang: Good idea. Will do.

src/protocols/displayport.stanza
```diff
@@ -0,0 +1,220 @@
...
+<DOC>
+public defn connect-displayport-lane (sw:Toleranced, ml:Double, cap:Instantiable|False, src:JITXObject, dst:JITXObject) :
+  inside pcb-module:
+    for i in indices(src.lane) do :
+      match(cap) :
+        (c:Instantiable) :
+          inst bl-cap : dp-coupler(c)
+          require tx1 : dual-pair from bl-cap
+          topo-net(src.lane[i], tx1.A)
+          topo-net(tx1.B, dst.lane[i])
```
callendorph, line 93: We should not change the duplication here. 
I don't think we fully understand the abstraction yet. 
There is clearly a repetitive case here that I think has additional features. As we discussed - ESD protection diodes are another case where there maybe a similar pattern. 

Let's think about how this might be better expressed. In the future, if/when we come up with a solution we can refactor.

src/protocols/displayport.stanza
```diff
@@ -0,0 +1,220 @@
...
+  inside pcb-module:
+    match(res, vdd, cap) :
+      (r:Instantiable, v:JITXObject, c:Instantiable) :
+        inst shunt-r : symmetric-shunt(r)
+        inst bl-cap : dp-coupler(c)
+        require tx1 : dual-pair from bl-cap
+        topo-net(src.aux, shunt-r.A)
+        topo-net(shunt-r.B, tx1.A)
+        topo-net(tx1.B, dst.aux)
+        topo-net(shunt-r.COMMON, v)
```
callendorph, line 126: Ya - there are a lot of args and connections here. 

This makes me want to do something like [Network](https://github.com/callendorph/etools/blob/master/src/jitx/Network.stanza) - but I don't think we know the abstraction yet.

src/protocols/displayport.stanza
bhusang: Thx.

src/protocols/displayport.stanza
```diff
@@ -0,0 +1,220 @@
...
+@param res Optional pull-up resistor to be added to the AUX pair before the capacitor. 
+Note that there are 2 resistors added, one on each of the P,N signals of the pair.
+@param vdd Power supply pin to be used for the optional pull-up resistors. Must be present if the resistors are used.
+@param cap Optional DC blocking capacitor to be inserted between the AUX pair. Note that
+there are 2 blocking caps inserted, one on each of the P,N signals of the pair.
+@param src This is assumed to be the source side of the link. This is expected to be a port of `Bundle` type `displayport`
+@param dst This is assumed to be the sink side of the link. This is expected to be a port of `Bundle` type `displayport`
+
+<DOC>
+public defn connect-displayport-aux (sw:Toleranced, ml:Double, res:Instantiable|False, vdd:JITXObject|False, cap:Instantiable|False, src:JITXObject, dst:JITXObject) :
```
callendorph, line 116: I think we should consider making a different polymorphic function here - or using named keyword arguments. 

This is going to get used as either: 

```
connect-displayport-aux(sw, ml, false, false, cap, src, dst) 
connect-displayport-aux(sw, ml, res, vdd, cap, src, dst) 

```

The `false` argument without any label or context is difficult to read. When I'm looking at this code without knowing the implementation - I will have no idea what the `false, false` is referring to. 

Having two variables that are dependent is also a non-optimal pattern. It is really easy to introduce bugs inadvertently. To wit - the case where the user passes `vdd/res` but not `cap` is not covered. 

There is also not much that is shared between these two cases that necessitates it all being in one function .

I would suggest creating: 

```
defn connect-displayport-aux (
  sw:Toleranced, ml:Double, 
  src:JITXObject, dst:JITXObject 
  -- 
  cap:Instantiable = ?,
  term:[Instantiable, JITXObject] = ?) :
 
```

The ` = ?` syntax forces a `Maybe<Instantiable>` or `Maybe<[Instantiable, JITXObject]>` 

You can then use the match with only two arguments which reduces the number of cases that you have to handle. 

It also forces the user to do: 

```
connect-displayport-aux(sw, ml,  src, dst, cap = bcap ) 
connect-displayport-aux(sw, ml, src, dst, cap = bcap, term = [r, vdd] ) 
```

So that now, there is no `false, false` and I know explicitly the arg names.

callendorph: The problem with `= ?` format is when you have to cascade it. 

You reference this function below in `connect-displayport` and using `= ?` in the parent and the child functions is not a great developer experience. You end up needing a `match` in the parent to deal with the different cases instead of deferring to the child.

callendorph: Because of that - the polymorphic approach might be a better option: 

```
defn connect-displayport-aux ( sw:Toleranced, ml:Double, src:JITXObject, dst:JITXObject, cap:Instantiable|False):
   ...

defn connect-displayport-aux ( sw:Toleranced, ml:Double, src:JITXObject, dst:JITXObject, cap:Instantiable|False, term:[Instantiable, JITXObject]|False) :

```
The second one can call the first one in the case where `term` is `False`.

src/si/couplers.stanza
```diff
@@ -95,4 +95,36 @@ public pcb-module symmetric-pull-downs (comp:Instantiable -- margin:Double = 0.0
...
+This function constructs a shunt module where two 2-pin components
+are instantiated with p[1] of both components adjacent to each other.
+The far pin (p[2]) on both components is connected to a `COMMON` terminal 
+which could be either a ground or a power supply net.
+
+@param comp 2-pin component like a resistors/capacitor with `p[1]` and `p[2]`
+@param margin Additional Y spacing distance between the two instantiated components.
+
+<DOC>
+public pcb-module symmetric-shunt (comp:Instantiable -- margin:Double = 0.0) :
```
callendorph, line 112: `symmetric-pull-downs` and `symmetric-shunt` are basically the same - you just changed `GND` to `COMMON` right ? 

If that is the case - then lets just rename `symmetric-pull-downs` and to `symmetric-shunt`. I like that name better and it is more versatile. 

The only place that `symmetric-pull-downs` is used is USB C converter and it isn't released - so I'll just rename there when this merges.

examples/protocols/displayport/displayport-board.stanza
bhusang: Done.

src/protocols/displayport.stanza
bhusang: Thanks -- a lot nicer.

src/si/couplers.stanza
bhusang: I didn't want to change a function name without asking 😄

src/protocols/displayport.stanza
bhusang: Yep -- food for thought.

src/protocols/displayport.stanza
bhusang: Is there a code change that I need to do for your comment?

callendorph (approved):
Looks good :+1: 

I think `pcie` is still using its own `pcie-board.stanza` file - we can modify that in a follow up commit.

PR #37, opened 2024-05-17
JITX-8584: improve spacing between pin names on opposite sides of box symbol

emspin:
- add minimum grid chunk to account for spacing between box and pin name when calculating total width

before: 

Screenshot 2024-05-16 at 6 25 01 PM

after: 

Screenshot 2024-05-16 at 6 23 56 PM


callendorph (approved):
:+1: Thanks @emspin

PR #38, opened 2024-05-17
Added SON package and USB protocol

callendorph:
This adds the SON (or DFN) land pattern package
It also adds the USB protocol definitions. 

@bhusang  - I need help with defining what the right set of values for `jsl/protocols/usb/USBVersion` should be and their corresponding skew, loss, and impedance values. 

I'm wondering if we should separate it into USB3 Gen1, USB3 Gen2, etc.


src/protocols/usb.stanza
```diff
@@ -0,0 +1,306 @@
...
+public pcb-enum jsl/protocols/usb/USBVersion:
+  USB2
+  USB3
+  USB4
+
+public defn usb-get-skew-loss (gen:jsl/protocols/usb/USBVersion) -> [Toleranced, Double]:
+  switch(gen):
+    USB2 : [0.0 +/- 3.75e-12, 2.0]
+    USB3 : [0.0 +/- 1.0e-12, 2.0]
+    USB4 : [0.0 +/- 1.0e-12, 2.0]
```
bhusang, line 153: These are ok for now -- likely very conservative from what I have seen.

bhusang (approved):

callendorph:
I'm working on a USB example like the pcie/displayport

callendorph:
I'm retracting and I'm going to introduce some new concepts and then provide the USB protocol in a better way.

PR #42, opened 2024-05-30
Added ability to name individual topo-net segments.

callendorph:
This allows you to name the nets of a topo net. This makes naming of labels in the schematic easier : 

```
  topo-net(
    usb2-src =>
    esd-data.A =>
    esd-data.B =>
    repeater.A =>
    repeater.B =>
    usb2-dst,
    [
      "TypeA-to-ESD"
      "ESD-underside"
      "ESD-to-Repeater"
      "Repeater-underside"
      "Repeater-to-TypeC"
    ]
  )

```

There are 5 segments and each segment gets a name.

### DIFF
--- a/src/si/helpers.stanza
+++ b/src/si/helpers.stanza
@@ -26,10 +26,17 @@ This function must be used in a `pcb-module` context.
 @param p1 Pin object
 @param p2 Pin object
 <DOC>
-public defn topo-net (p1:JITXObject, p2:JITXObject):
+public defn topo-net (p1:JITXObject, p2:JITXObject -- name?:Maybe<String> = None()):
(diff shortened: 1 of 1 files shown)


bhusang (approved):
Putting the names after the connections seems backwards to me 😸. But I suspect that is the only option right now unless we change the macro definition?

callendorph:
I want to refactor this to have a single base name and then have the `topo-net` function create names like `base.<ID>` so that we can have a easier way to name and have consistency between names. What is important is that the user can some consistent way to know what nets are a part of what topology.

callendorph:
I'm retracting this as we have a better method in the works.

PR #43, opened 2024-05-30
Updated the diff-impedance for the examples to be parametric.

callendorph:
Leverage more of the API for the protocols.


bhusang (approved):
Nice.

PR #46, opened 2024-06-05
Added `gates` bundles for implementing logic gate supports/require

callendorph:
This adds a base bundle type and a helper function for creating the supports statements for a N-input bundle.

This also adds a simple example that demonstrates the concept:

Screenshot 2024-06-04 at 9 49 21 PM


bhusang (approved):

callendorph:
OK - I was able to simplify a bit. One thing I found to still be annoying was this: 

```
  val gate-symbs = for i in 0 to num-circuits seq:
    i => create-symbol $ ANDGateSymbol(
      in-ref = #R(AND[i].in)
      out-ref = #R(AND[i].out)
    )
```

I had to use the `#R` Ref short hand here instead of using `self.AND[i].out` directly because the symbol expects a Ref. 

I think we should consider making the Symbol implemenations tolerant of either `Ref` or `JITXObject`. The `JITXObject` variant we would expect to be a port. But then it could call `ref()` on that port and get the `Ref` object that we need for the symbol. 

I checked in the example and if I do `ref(self.AND[i].out)` it converts to `AND[i].out`. So I think this might be a way to simplify things to: 

```
  val gate-symbs = for i in 0 to num-circuits seq:
    i => create-symbol $ ANDGateSymbol(
      in-ref = self.AND[i].in,
      out-ref = self.AND[i].out
    )
```

Which at least requires less magic syntax.

PR #48, opened 2024-06-06
Ethernet Protocol

callendorph:
Initial structure for the ethernet protocol. 

I need to add: 

1.  Examples 
2.  Constraints for GMII, MII, and RMII
     1.  We may want to reorganize the bundles to make them "Lane" like so that we can do a similar pattern to what USB and RGMII do. 
4.  Media Dependent Interface (MDI) for the diff-pairs going from the PHY to the connector
5. SGMII ?


bhusang (approved):

callendorph:
I've added the 1000Base-T MDI so that we have an example of an MDI 
I've updated the structure of the repositories so that we have a MII and MDI sub directory for the ethernet protocols
I've added an example using RGMII and 1000Base-T. 

Note - The example only works on develop right now. It throws an error on JITX 3.9.0 which is unfortunate.

@bhusang  - We should use this structure and layer your 10G-KR MDI on top of it.

PR #49, opened 2024-06-06
Add ethernet-10G-KR

bhusang:
This PR also has a number of ancillary changes across JSL for minor issues, etc.


src/protocols/ethernet-10G-KR.stanza
```diff
@@ -0,0 +1,146 @@
...
+in order for the overall skew and loss values to be respected.
+@param cap-dst-src Optional DC blocking capacitor to be inserted between the dst.TX and src.RX signal ends. This capacitor
+covers the reverse direction connection between dst.TX and src.RX. If the second component was a connector, then typically
+no capacitor is inserted between dst.TX and src.RX (cap-dst-src = false).
+@param src This is expected to be a port of `Bundle` type `ethernet-10G-KR`
+@param dst This is expected to be a port of `Bundle` type `ethernet-10G-KR`
+@param null-modem If set to true, connects src.TX through the optional cap-src-dst to dst.RX and dst.TX through the optional cap-dst-src to src.RX. 
+If set to false, connects src.TX through the optional cap-src-dst to dst.TX and dst.RX through the optional cap-dst-src to src.RX
+<DOC>
+public defn connect-10G-KR (cons, cap-src-dst:Instantiable|False, cap-dst-src:Instantiable|False, src:JITXObject, dst:JITXObject, null-modem:True|False) :
```
callendorph, line 67: `cons` is lacking a type. It should be `SI-Constraint`

You might want to consider using a keyword arg for `null-modem` as a random true/false at the end of the function is going to be hard to read. 

---

I don't agree with this approach to defining constraints on topologies.

1.  Repetition - notice how every block of the match has very similar statements
2.  Has many arguments - for me, this is a leading indicator that the structure is wrong.
3. Not Extensible - it covers cases that may never happen, but also is not flexible enough to give the user options. This makes this function very brittle. 

I know we want to make it seem like magic, but this isn't magic to me. This is akin to a reference design. It is one instance of making this work - it is not an expressive system from which I can make any 10G-KR design. 

I also know we need to get this out. I think we are still in the alpha stage and so in my opinion we can make API changes as we want. This function also doesn't stop me or the end user from using what I think is the "right" way to design this - so I think we should move forward in service of haste.

src/pin-assignment.stanza
```diff
@@ -286,7 +286,7 @@ Swizzle a Bus Bundle
 
 This constructs a supports statement for a bus bundle.
 A bus bundle is a bundle with one PortArray pin definition.
-The support statement then constructs a cross-bar for the
+The support statement then constructs a cross-bar for the ?
```
callendorph, line 289: Ooops - I'll work on fixing that. Thanks!

src/pin-assignment.stanza
```diff
@@ -27,7 +27,7 @@ public defn check-bundle (exp:Bundle, b-set:JITXObject ...) :
     match(port-type(p1)):
       (obs:Bundle):
         if obs != exp:
-          throw $ ArgumentError("Expected Bundle '%_' - Observed Bundle '%_'" % [name(obs), name(exp)])
+          throw $ ArgumentError("Expected Bundle '%_' - Observed Bundle '%_'" % [name(exp), name(obs)])
```
callendorph, line 30: Good catch

callendorph (approved):
Let's figure out how to rebase this on my ethernet PR that sets up the structure of MII and MDI and then merge this.

PR #50, opened 2024-06-07
Circuit Pool 

callendorph:
This PR introduces the concept of a circuit pool.

This is a generalization of the approach I used in the USB A/C converter for the ESD protection. 
I think it will be useful for many other applications as well. Basically any circuit where we need to use multiple multi-channel packages (Quad AND, Quad / Dual Op Amps, Dual Comparator, etc), the Circuit Pool can provide a convenient way to apply pin assignment across a pool of components. 

Here is a capture from the example: 
Screenshot 2024-06-07 at 1 45 06 PM

In the example, I create an ESD pass through protector component and then create a Circuit Pool using it as the template. This allows any diff-pair ESD protection channel to be used for any of the 1000Base-T MDI pairs. 

I've also refactored the `pin-assignment` code a bit.


bhusang (approved):

bhusang:
LGTM but github is complaining about a conflict?

callendorph:
Ya - let me fix that.

callendorph:
Fixing this is a pain. I'm going to close this and re-submit a new PR based off main.

PR #54, opened 2024-06-10
Add Reference Designators for Package Generators

callendorph:
This adds a new feature to generate an initial reference designator placement in landpattern generators. 

I've added an example for the SON package. 
I've added some refactors to reduce code duplication.

Examples should now show this:

Screenshot 2024-06-10 at 1 46 57 PM

With a U1 or similar ref-des


callendorph:
I want to add a more global default configuration for ref des size

bhusang (approved):

PR #56, opened 2024-06-10
Add ARINC818 protocol

bhusang:


examples/protocols/arinc818/ARINC818-main.stanza
```diff
@@ -0,0 +1,47 @@
+#use-added-syntax(jitx)
+defpackage jsl/examples/protocols/sata/SATA-main:
```
callendorph: Package name is SATA instead of ARINC818

examples/protocols/arinc818/ARINC818-src.stanza
```diff
@@ -0,0 +1,126 @@
...
+      supports arinc :
+        require dp: diff-pair from ctrl
+          if invert :
+            arinc.TX.P => dp.N
+            arinc.TX.N => dp.P
+          else :
+            arinc.TX.P => dp.P
+            arinc.TX.N => dp.N
+
+pcb-landpattern sata-lp :
```
callendorph: Why are we reusing `sata` names in these files ?

bhusang: This is an old file!

bhusang: The version that I just pushed has no sata names in the file.

callendorph: I see - so Github's visualizer is bugging out ?

bhusang: Yeah, I think that you have stale data in the UI.

src/protocols/arinc818.stanza
```diff
@@ -0,0 +1,160 @@
...
+The Data Bundle consists of 1 differential pair (1 RX or TX)
+@member TX/RX for the ARINC818 bundle (consists of either a TX or RX diff pair)
+<DOC>
+
+public pcb-bundle ARINC818 (type:ARINC818Type) :
+  name = "ARINC818"
+  description = "ARINC 818 AVDB protocol"
+  switch(type) :
+    TX : make-port(`TX, diff-pair)
+    RX : make-port(`RX, diff-pair)
```
callendorph, line 67: For other unidirectional standards like DisplayPort, we don't differentiate between transmit and receive. 

It feels like this attempting to implement a Cardinality style implementation. Ie - this is a output, this is an input, kind of thing.  

Side note - is the ARINC818 really just a diff-pair ? Do we have a copy of the spec ?

src/protocols/arinc818.stanza
```diff
@@ -0,0 +1,160 @@
...
+
+doc: \<DOC>
+@brief Check the presence of a TX and RX version of the ARINC818 bundle
+
+This function checks the presence of the TX and RX pins on the ARINC818 bundle.
+
+@param p-tx The TX pin
+@param p-rx The RX pin
+<DOC>
+public defn check-correct-pins (p-tx:JITXObject p-rx:JITXObject):
```
callendorph, line 98: We have a `check-single-pin` already:

https://github.com/JITx-Inc/jsl/blob/6c9232ff401c2e14cc12158be732462d5221a37b/src/pin-assignment/checks.stanza#L9

bhusang: Yep, tried to use that first. But the error message would not make sense in the context that you did not connect the TX to RX but instead connected RX to TX. The reason to make this directional is that some implementations are optical and in that case, you need some more directionality?

PR #59, opened 2024-06-13
added optional PRSNT# pin

bhusang:


src/protocols/pcie.stanza
```diff
@@ -70,22 +70,26 @@ public defn PCIe-enum-to-int (en:PCIeWidth) -> Int:
...
+@param opt-pins Optional pin support where the value is an enum from jsl/bundles/pcie/PCIePins
 <DOC>
 
 ; for the future
 ; P/N swapping is ok per pair
 ; all lanes being swapped is ok as well
 ; this should be handled at the component level
 
-public pcb-bundle pcie (lanes:Int) :
+public pcb-bundle pcie (lanes:Int -- opt-pins:Collection<PCIePins>|False = false) :
```
callendorph, line 89: Default arguments in `pcb-bundle` definitions don't work. 

Can we please follow the pattern we have established in other functions - for example: 

```
public defn smbus (pins:SMBUSPins ...) :
  smbus-b(pins)
```

Where we create a separate function that wraps the bundle definition and allows the user to just use `smbus()` when there are no options needed.

callendorph: This comment applies to all bundles that use the optional pins.

callendorph (approved):
Looks good - Please see the comment regarding default args

src/protocols/pcie.stanza
```diff
@@ -105,15 +109,22 @@ public pcb-bundle pcie-data (lanes:Int) :
 doc: \<DOC>
 @brief PCI-e Control bundle
 PCI-e is a serial communication supporting high speed data links
+@param pins An optional set of pins to be implemented in the control bundle
```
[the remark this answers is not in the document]
bhusang: Sounds good.

src/protocols/pcie.stanza
bhusang: I tried the default argument in the pcie bundle by running the pcie example. What aspect of the default argument should not work?

bhusang: I will modify the bundle to match.

callendorph: You pass an argument in. 
Have you tried running it without an `opt-pins` argument ?

bhusang: Yes, I did. The pcie example code did both (with and without the opt-pins argument). Seemed to work.

callendorph: I'll have to check again. I've structured all the others this way because I ran into issues using the advanced keyword arguments and defaults on `pcb-bundle`. 

I wonder if it was the variadic that was not working, not the keyword args.

PR #62, opened 2024-06-14
PROD-132/Jsl/layerstack access functions

lchao-jitx:
Support access functions for Layerstack:
  - get layer by index function:  `test-stack[0]`
  - get conductor layer by index: `conductors(test-stack)[0]`
  - set-name function: `set-name(test-stack[0] One("cu3-new"))`
  - set function to change a layer by index: `test-stack[0] = new-stack`
  - allow arguments of `Copper()` functions to be either weight or thickness.
  
  I created the ticket PROD-132 for this issue, but forgot to include it in the branch name.

### DIFF
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -10,21 +10,18 @@ defpackage jsl/layerstack:
 doc: \<DOC>
 Base class for Defining the Layer Material
 <DOC>
-public-when(TESTING) defstruct LayerMaterial <: Hashable & Equalable :
+public defstruct LayerMaterial <: Hashable & Equalable :
   name:Maybe<String>
   type:MaterialType
   description:Maybe<String>
   material-def:Maybe<JITXDef> with: (
     setter => set-material-def
   )
+with:
+  equalable => true
+  hashable => true
 
 defmulti create-material (x:LayerMaterial) -> JITXDef
-
-defmethod equal? (ma:LayerMaterial, mb:LayerMaterial) :
-  name(ma) == name(mb) and
-  type(ma) == type(mb) and
-  material-def(ma) == material-def(mb)
-
 
 ; I need these accessors because the `name` and `type` tokens
 ;  will get consumed by the macro before they can call the functions.
@@ -36,7 +33,7 @@ defn make-name (x:LayerMaterial):
       (_:None): false
       (v:One<String>): name = value(v)
 
-public-when(TESTING) defn get-type (x:LayerMaterial) -> MaterialType:
+public defn get-type (x:LayerMaterial) -> MaterialType:
   type(x)
 
 defn make-description (x:LayerMaterial) :
@@ -221,7 +218,7 @@ public val SoldermaskMaterial = DielectricMaterial(
 )
 
 public defstruct LayerSpec <: Equalable :
-  name:Maybe<String> with: (setter => set-name)
+  name:Maybe<String> with: (setter => set-name?)
   description:Maybe<String>
   material:LayerMaterial
   thickness:Double with: (
@@ -230,11 +227,10 @@ public defstruct LayerSpec <: Equalable :
 with:
   constructor => #LayerSpec
   printer => true
+  equalable => true
 
-defmethod equal? (la:LayerSpec, lb:LayerSpec) :
-  name(la) == name(lb) and
-  material(la) == material(lb) and
-  thickness(la) == thickness(lb)
+public defn set-name (l:LayerSpec, name:String) :
+  set-name?(l One(name))
 
 public defn LayerSpec (
   --
@@ -247,37 +243,25 @@ public defn LayerSpec (
 
 doc: \<DOC>
 Create a Copper layer
-  Usage: Copper(name, thickness) : specify the thickness directly
-         Copper(name, weight, weight? = true) : specify weight in order to compute thickness
-         Copper(weight, AluminumMaterial, weight? = true) : without name
+  Usage: Copper(0.3) : specify thickness 
+         Copper(0.3, name = "Cu") : specify thickness and name
+         Copper(0.3, AluminumMaterial, name = "Cu") : specify thickness, material and name
+@param thickness : layer thickness in mm
+@param material : optional positional argument
+@param name : optional named argument
 <DOC>
-public defn Copper (name:String, thickness:Double, material:ConductorMaterial = CopperMaterial
-    -- weight?:True|False = false) -> LayerSpec :
-  if weight? :
-    val weight = thickness
-    LayerSpec(name = name, thickness = compute-thickness(material, weight), material = material)
-  else : 
-    LayerSpec(name = name, thickness = thickness, material = material)
-
-public defn Copper (thickness:Double, material:ConductorMaterial = CopperMaterial
-    -- weight?:True|False = false) -> LayerSpec :
-  if weight? :
-    val weight = thickness
-    LayerSpec(thickness = compute-thickness(material, weight), material = material)
-  else :
-    LayerSpec(thickness = thickness, material = material)
-
-public defn FR4 (name:String, thickness:Double, material:DielectricMaterial = FR4-Material) -> LayerSpec :
-  LayerSpec(name = name, material = material, thickness = thickness)
-
-public defn FR4 (thickness:Double, material:DielectricMaterial = FR4-Material) -> LayerSpec :
-  LayerSpec(material = material, thickness = thickness)
-
-public defn Soldermask (name:String, thickness:Double, material:DielectricMaterial = SoldermaskMaterial) -> LayerSpec :
-  LayerSpec(name = name, material = material, thickness = thickness)
-
-public defn Soldermask (thickness:Double, material:DielectricMaterial = SoldermaskMaterial) -> LayerSpec :
-  LayerSpec(material = material, thickness = thickness)
+#for (
+  func-name in [Copper FR4 Soldermask]
+  material-type in [ConductorMaterial DielectricMaterial DielectricMaterial ]
+  material-default in [CopperMaterial FR4-Material SoldermaskMaterial]
+) :
+  public defn func-name (thickness:Double, material:material-type = material-default
+      -- name:String = ?) -> LayerSpec :
+    match(name):
+      (name:One) :
+        LayerSpec(name = value(name), thickness = thickness, material = material)
+      (_):
+        LayerSpec(thickness = thickness, material = material)
 
 defn get-name (x:LayerSpec) -> Maybe<String> :
   name(x)
@@ -319,20 +303,9 @@ public defn LayerStack (
 defmethod write (o:OutputStream, s:LayerStack) :
   println(o, "LayerStack: %_" % [value?(name(s))])
   println(o, "  Layers: %_ copper layers %_ total layers" % [get-conductor-count(s) length(layers(s))])
-  for [idx, l] in enumerate(layers(s)) do:
+  for (l in layers(s), idx in 0 to false) do :
     println(o, "    %_ : %_ | %_ | %_" %
       [idx, value?(name(l)) get-type(material(l)) thickness(l)])
-
-doc:\<DOC>
-  Usage: for [idx, line] in enumerate(lines) do: 
-           println(“Line %: %_” % [idx, line])
-<DOC>
-defn enumerate<?T> (s:Seqable<?T>) -> Seq<[Int, T]> :
-  var i = 0
-  for x in s seq :
-    next where:
-      val next = [i, x]
-      i = i + 1
 
 doc:\<DOC>
   Get a layer by index
@@ -446,13 +419,13 @@ layers, and 2 more copper layers. This constructs a symmetric stackup.
 ```
 val stack = LayerStack(name = "6-Layer Symmetric Stackup")
 
-val copper-1oz = Copper("cu", 1.0, weight = true)
-val core-FR4 = FR4("core", 1.0)
-val prepreg-FR4 = FR4("prepreg", 0.5)
+val copper-1mm = Copper(1.0, name = "cu")
+val core-FR4 = FR4(1.0, name = "core")
+val prepreg-FR4 = FR4(0.5, name = "prepreg")
 
-add-symmetric(copper-1oz, core-FR4,
-  add-symmetric(copper-1oz, prepreg-FR4,
-    add-symmetric(copper-1oz, core-FR4, stack)
+add-symmetric(copper-1mm, core-FR4,
+  add-symmetric(copper-1mm, prepreg-FR4,
+    add-symmetric(copper-1mm, core-FR4, stack)
   )
 )
 ```
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -236,20 +236,25 @@ public defn LayerSpec (
 
 doc: \<DOC>
 Create a Copper layer
+  Usage: Copper(name, thickness) : specify the thickness directly
+         Copper(name, weight, weight? = true) : specify weight in order to compute thickness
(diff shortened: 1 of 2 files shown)


src/layerstack.stanza
```diff
@@ -10,7 +10,7 @@ defpackage jsl/layerstack:
 doc: \<DOC>
 Base class for Defining the Layer Material
 <DOC>
-defstruct LayerMaterial <: Hashable:
+public-when(TESTING) defstruct LayerMaterial <: Hashable & Equalable :
```
callendorph, line 13: Let's just make this `public`

src/layerstack.stanza
```diff
@@ -20,6 +20,12 @@ defstruct LayerMaterial <: Hashable:
 
 defmulti create-material (x:LayerMaterial) -> JITXDef
 
+defmethod equal? (ma:LayerMaterial, mb:LayerMaterial) :
```
callendorph, line 23: The `JITXDef` types are `Equalable` - so I think you should be able to just add `equalable => true` to the `defstruct LayerMaterial` definition.  For example: 

```
public-when(TESTING) defstruct LayerMaterial <: Hashable & Equalable :
  name:Maybe<String>
  type:MaterialType
  description:Maybe<String>
  material-def:Maybe<JITXDef> with: (
    setter => set-material-def
  )
with:
  equalable => true
  ; If we want to make this public - then we also need this.
  hashable => true
```

Further, you are missing the `description` so this `equal?` method will not work in all cases. 

We should write a test for this to be sure.

src/layerstack.stanza
```diff
@@ -30,7 +36,7 @@ defn make-name (x:LayerMaterial):
       (_:None): false
       (v:One<String>): name = value(v)
 
-defn get-type (x:LayerMaterial) -> MaterialType:
+public-when(TESTING) defn get-type (x:LayerMaterial) -> MaterialType:
```
callendorph, line 39: Same comment - Lets just make this public.

src/layerstack.stanza
```diff
@@ -225,6 +231,11 @@ with:
   constructor => #LayerSpec
   printer => true
 
+defmethod equal? (la:LayerSpec, lb:LayerSpec) :
```
callendorph, line 234: Same comment here - Lets not make an explicit `equal?` lets use the `equalable => true` in the defstruct definition.

src/layerstack.stanza
```diff
@@ -214,8 +220,8 @@ public val SoldermaskMaterial = DielectricMaterial(
   is-soldermask = true
 )
 
-public defstruct LayerSpec:
-  name:Maybe<String>
+public defstruct LayerSpec <: Equalable :
+  name:Maybe<String> with: (setter => set-name)
```
callendorph, line 224: Let's make this internal setter `set-name?` where `?` indicates maybe. 

Then lets create a `defn set-name` without the `?` that takes just a `String` argument and converts it to `Maybe<String>` with `One`.

That way the user facing interface is simpler.

src/layerstack.stanza
```diff
@@ -236,12 +247,25 @@ public defn LayerSpec (
...
+  Usage: Copper(name, thickness) : specify the thickness directly
+         Copper(name, weight, weight? = true) : specify weight in order to compute thickness
+         Copper(weight, AluminumMaterial, weight? = true) : without name
 <DOC>
-public defn Copper (name:String, weight:Double, material:ConductorMaterial = CopperMaterial) -> LayerSpec :
-  LayerSpec(name = name, thickness = compute-thickness(material, weight), material = material)
-
-public defn Copper (weight:Double, material:ConductorMaterial = CopperMaterial) -> LayerSpec :
-  LayerSpec(thickness = compute-thickness(material, weight), material = material)
+public defn Copper (name:String, thickness:Double, material:ConductorMaterial = CopperMaterial
```
callendorph, line 254: I'm  going to propose that we simplify this and remove the `weight` argument completely:

```
public defn Copper (thickness:Double, material:ConductorMaterial = CopperMaterial -- name:String = ?) -> LayerSpec:
   ...

```

The `weight` argument doesn't work well because there does not seem to be an established, consistent density for the copper foil applied to boards. Depending on what value is chosen for the density we get different thicknesses. I think we should simplify to just a thickness in mm.

src/layerstack.stanza
```diff
@@ -278,7 +302,7 @@ public defstruct LayerStack :
   description:Maybe<String>
 
   layers:Vector<LayerSpec> with: (
-    updater => sub-layers
+    setter => set-layers
```
callendorph, line 305: Why are we making this change ?

What is driving a need for mutability here ?

src/layerstack.stanza
```diff
@@ -292,6 +316,49 @@ public defn LayerStack (
...
+doc:\<DOC>
+  Usage: for [idx, line] in enumerate(lines) do: 
+           println(“Line %: %_” % [idx, line])
+<DOC>
+defn enumerate<?T> (s:Seqable<?T>) -> Seq<[Int, T]> :
+  var i = 0
+  for x in s seq :
+    next where:
+      val next = [i, x]
+      i = i + 1
```
callendorph, line 335: Can't you also do this with:

```
for (x in s, i in 0 to false) seq:
  [i, x]
```

Isn't this simpler and avoids a mutable variable.

src/layerstack.stanza
```diff
@@ -292,6 +316,49 @@ public defn LayerStack (
...
+    val new-layers = to-tuple $ cat-all $
+      [layers(stack)[0 to idx] [layer] layers(stack)[(idx + 1) to false]]
+  ;<A>
+
+doc:\<DOC>
+  Get the idx-th conductor layer
+  Usage: conductors(stack)[idx]
+<DOC>
+public defn conductors (stack:LayerStack) -> Tuple<LayerSpec> :
+  to-tuple $ filter({material(_) is ConductorMaterial} layers(stack))
```
callendorph, line 361: This is good :+1:

src/layerstack.stanza
```diff
@@ -292,6 +316,49 @@ public defn LayerStack (
...
+  Usage: stack[idx]
+<DOC>
+public defn get (stack:LayerStack, idx:Int) -> LayerSpec :
+  layers(stack)[idx]
+
+doc:\<DOC>
+  Modify a layer by index
+  Usage: stack[idx] - new-layer
+<DOC>
+public defn set (stack:LayerStack, idx:Int, layer:LayerSpec) -> False :
```
callendorph, line 348: Why do we need this ? 
What is the use case that we want to add a `set` operator here ?

callendorph (changes requested):
Hi @lchao-jitx  - This coming along nicely. 

I've made a few comments. Please let me know if you have any questions.

src/layerstack.stanza
callendorph: I don't think that is true.

lchao-jitx: Adding just `true => equalable` works.

This `equal?` function is already tested since I need it to test `state[0]` etc. For example,
```
deftest(layerstack) test-get :
  #EXPECT(test-stack[0] == copper-3oz)
  #EXPECT(test-stack[1] == core-FR4-2)
  #EXPECT(test-stack[2] == copper-2oz)
  #EXPECT(test-stack[3] == prepreg-FR4)
  ```

File: src/layerstack.stanza (hunk 5 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:224]: Should we change the field name `name` to `name?` to be consistent?
Then, add a function `name` to access it as a String or "" if empty.

Comment (callendorph) [reply src/layerstack.stanza:224]: I don't have an issue with that proposal as long as we stay consistent. So anywhere in this file (or ideally, anywhere in JSL) where we use a `Maybe` needs to use a `?` suffix.

File: src/layerstack.stanza (hunk 6 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:254]: Replacing them with this:
```
#for (
  func-name = [Copper FR4 Soldermask]
  material-type = [CopperMaterial DielectricMaterial DielectricMaterial ]
  material-default = [CopperMaterial FR4-Material SoldermaskMaterial]
) :
  public defn func-name (thickness:Double, material:material-type = material-default
    -- name:String = ?) -> LayerSpec :
  LayerSpec(name = name, thickness = thickness, material = material)
```

File: src/layerstack.stanza (hunk 7 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:305]: To support the `set` function (since we have the `get` function):

```
; Test: test-stack[0] = new-stack
deftest(layerstack) test-set :
  #EXPECT(test-stack[0] == copper-3oz)
  test-stack[0] = copper-1oz
  #EXPECT(test-stack[0] == copper-1oz)

  #EXPECT(test-stack[1] == core-FR4-2)
  test-stack[1] = core-FR4-1
  #EXPECT(test-stack[1] == core-FR4-1)
```

Comment (callendorph) [reply src/layerstack.stanza:305]: I would prefer we not implement this as a setter. The `sub-*` method allows this to be a transformation but maintains immutability. 

Additionally - we can very easily add this without breaking the ABI if we find that we need it later.

File: src/layerstack.stanza (hunk 9 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:361]: Come to think of it: `layers(stack)` returns a `Vector`.
But `conductors(stack)` would return a tuple.
This is not consistent. But maybe it does not matter much.

`layers` in `LayerStack` only needs to be a vector internally.

File: src/layerstack.stanza (hunk 10 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:348]: I just thought we might need to support `set` since we support `get`.
This would replace a layer in the stack by another layer.
If you don't think that's useful, happy to remove it.

```
deftest(layerstack) test-set :
  #EXPECT(test-stack[0] == copper-3oz)
  test-stack[0] = copper-1oz
  #EXPECT(test-stack[0] == copper-1oz)

  #EXPECT(test-stack[1] == core-FR4-2)
  test-stack[1] = core-FR4-1
  #EXPECT(test-stack[1] == core-FR4-1)
```

File: src/layerstack.stanza (hunk 9 above)
Comment (callendorph) [reply src/layerstack.stanza:361]: I think immutability of the `Tuple` here is a good thing.

File: src/layerstack.stanza (hunk 10 above)
Comment (callendorph) [reply src/layerstack.stanza:348]: I want to avoid this kind of mutability. The goal is that the layer stack is declarative as much as possible.

File: src/layerstack.stanza (hunk 5 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:224]: On second thought, not changing the field name `name` to `name?`.
There are many other fields with the `Maybe` type. To be consistent, we would need to change all of them to `field?`. Maybe you don't want so many changes.

File: src/layerstack.stanza (hunk 10 above)
Comment (lchao-jitx) [reply src/layerstack.stanza:348]: I'll remove the set function then.

**Comment by lchao-jitx** [comment]
I added the `write` method for LayerStack for testing. Do you want to make it available in `src/layerstack.stanza`?
If so, I probably should find out how to print a nicer table format.
If not, I won't spend time on that.
It currently prints this:
```
LayerStack: My Crazy Stack
  Layers: 6 conductor layers 11 total layers
    0 : cu3 | Conductor | 3.0
    1 : core2 | Dielectric | 2.0
    2 : cu2 | Conductor | 2.0
    3 : prepreg | Dielectric | 0.5
    4 : cu1 | Conductor | 1.0
    5 : core1 | Dielectric | 1.0
    6 : cu1 | Conductor | 1.0
    7 : prepreg | Dielectric | 0.5
    8 : cu2 | Conductor | 2.0
    9 : core2 | Dielectric | 2.0
    10 : cu3 | Conductor | 3.0
```

File: src/layerstack.stanza [hunk 11]
```diff
@@ -236,24 +243,25 @@ public defn LayerSpec (
 
 doc: \<DOC>
 Create a Copper layer
+  Usage: Copper(0.3) : specify thickness 
+         Copper(0.3, name = "Cu") : specify thickness and name
+         Copper(0.3, AluminumMaterial, name = "Cu") : specify thickness, material and name
```
callendorph, line 248: In the third example: 

```
Copper(0.3, AluminumMaterial, name = "Cu")
```

We should probably use `name = "PWR"` instead of `Cu` which is short hand for Copper.

src/layerstack.stanza
```diff
@@ -236,24 +243,25 @@ public defn LayerSpec (
 
 doc: \<DOC>
 Create a Copper layer
```
callendorph, line 245: Documenting functions defined in a `#for` like this is difficult. I don't think there is very good support for it yet.

src/layerstack.stanza
```diff
@@ -278,7 +286,7 @@ public defstruct LayerStack :
   description:Maybe<String>
 
   layers:Vector<LayerSpec> with: (
-    updater => sub-layers
+    setter => sub-layers
```
callendorph, line 289: You renamed it but didn't convert it back to `updater` here. You have left it as a `setter`.

lchao-jitx: Oops!

src/layerstack.stanza
```diff
@@ -236,27 +245,36 @@ public defn LayerSpec (
...
+    (_):
+      LayerSpec(thickness = thickness, material = material)
+
+public defn Soldermask (thickness:Double, material:DielectricMaterial = SoldermaskMaterial
+    -- name:String = ?) -> LayerSpec :
+  match(name):
+    (name:One) :
+      LayerSpec(name = value(name), thickness = thickness, material = material)
+    (_):
+      LayerSpec(thickness = thickness, material = material)
```
callendorph, line 277: I understand that you did this with a `#for` and so there wasn't duplication but not that it is unrolled there is duplication. 

Can we please refactor these methods to use the `#LayerSpec` underlying constructor so that we can avoid the need to have two versions of the constructor call ? 

```
public defn Soldermask (thickness:Double, material:DielectricMaterial = SoldermaskMaterial
    -- name:String = ?) -> LayerSpec :
      #LayerSpec(name, None(), material, thickness)
```

Can we also put another optional keyword for `description:String = ?` so that the user can override that if they care to ?

src/layerstack.stanza
callendorph: This is still outstanding, I believe.

tests/layerstack.stanza
```diff
@@ -6,17 +6,18 @@ defpackage jsl/tests/layerstack:
...
   import jsl/tests/utils
+  import jsl/errors
 
 deftest(layerstack) test-basic-symmetric:
   val stack = LayerStack(name = "Carl's Crazy Stack")
 
-  val copper-1oz = Copper("cu", 1.0)
-  val core-FR4 = FR4("core", 1.0)
-  val prepreg-FR4 = FR4("prepreg", 0.5)
+  val copper-1mm = Copper(1.0, name = "cu")
```
callendorph, line 14: Same comment - `1mm` -> `35um`

tests/layerstack.stanza
```diff
@@ -154,3 +150,78 @@ deftest(layerstack) test-with-soldermask:
...
+ 4 - copper-1oz
+ 5 - core-FR4-1
+ 6 - copper-1oz
+ 7 - prepreg-FR4
+ 8 - copper-2oz
+ 9 - core-FR4-2
+ 10 - copper-3oz
+;<note>
+val test-stack = LayerStack(name = "My Crazy Stack")
+val copper-1mm = Copper(1.0, name = "cu1")
```
callendorph, line 169: We should use :

`1mm` -> `35um` 
`2mm` -> `70um` 
`3mm` -> `105um` 

For more realistic values.

callendorph:
I've pushed docs updates.

callendorph (approved):
:+1: Thanks @lchao-jitx !

callendorph:
I've run tests locally for the layer stack and they pass.

PR #63, opened 2024-06-14
added @member, more SPI pins, restructured DP with a new connector

bhusang:


src/bundles/comms.stanza
```diff
@@ -111,10 +111,23 @@ public pcb-bundle can-logical :
...
+SPI interfaces automatically, the constituent pins need to be present on all
+objects to be connected. 
+<DOC>
+
 public pcb-enum jsl/bundles/comms/SPIPins :
   SPI-CS
   SPI-MISO
   SPI-MOSI
+  SPI-COPI
+  SPI-CIPO
```
callendorph, line 126: `COPI` and `MOSI` are synonymous.
Same with `CIPO` and `MISO`. 

I don't think we should have two bundle definitions for the same thing with different names. 

If we want to use `CIPO/COPI` instead of `MISO/MOSI` - thats fine. Lets replace them - but lets not do both.

src/bundles/comms.stanza
```diff
@@ -111,10 +111,23 @@ public pcb-bundle can-logical :
...
+  SPI-RDY
+  SPI-SDO
+  SPI-SDI
```
callendorph, line 129: Same comment - I think having multiple different ways to express the same bundle is an anti-pattern. Lets please remove these or create a different bundle type.

The reason is because of our "quasi-inheritance" bundle mechanism. If we have several different components with bundles that all have different names, that mechanism will never work correctly.

src/bundles/comms.stanza
```diff
@@ -111,10 +111,23 @@ public pcb-bundle can-logical :
...
+objects to be connected. 
+<DOC>
+
 public pcb-enum jsl/bundles/comms/SPIPins :
   SPI-CS
   SPI-MISO
   SPI-MOSI
+  SPI-COPI
+  SPI-CIPO
+  SPI-RDY
```
callendorph, line 127: This signal is not standard SPI.  It is a custom extension for whichever MCU we found it in - right ? 

I think this is another case for the concept of inheritance in bundles. 

Something feels wrong about putting this here. I can't put my finger on it. But if I had bundle inheritance - I would definitely not put this pin definition here.

The reason to contemplate it at all is because of the complexity of this bundle. There are several different configuration variants - the only constant is `SCLK`.

The only other option I can think to do is to create helper "generator" functions that construct the internals of an SPI bundle consistently with options and then allow the user to add their custom port definitions in their derived SPI bundles.

@CuppoJava - FYI

src/protocols/displayport.stanza
```diff
@@ -54,16 +57,38 @@ doc: \<DOC>
...
+@member lane Four differential pairs for data transfer
+@member hpd Hot plug detect
 <DOC>
 
 public pcb-bundle displayport :
   name = "DisplayPort"
   description = "DisplayPort Digital Display Interface"
   port aux  : diff-pair
-  port lane : diff-pair[4]
+  port lane : diff-pair[DISPLAYPORT_NUM_LANES]
```
callendorph, line 70: We discussed that there is a miniature display port with fewer lanes - correct ? 

Should this not be an argument with default value ?

src/protocols/displayport.stanza
```diff
@@ -54,16 +57,38 @@ doc: \<DOC>
...
   port hpd  : pin ; hot plug detect
```
callendorph, line 71: Is this part of the standard high speed connections for the transport bundle - or should this be in the connector bundle ? 

Is it typical for SoC peripherals for Display port to have an explicit pin for the HPD ? Or is it usually just a GPIO ? 

If it is just a GPIO - then we should put it in the connector bundle.

src/protocols/usb.stanza
```diff
@@ -100,6 +107,8 @@ This bundle type is usually used to construct the ports on MCUs,
 ESD devices, Muxes, etc in the USB 3/4 applications.
 
 @param lane-cnt Number of superspeed lanes in this bundle. Default is 1.
+@member data Differential pair for data
```
callendorph, line 110: USB2.0 Data

callendorph (approved):
:+1: I made several comments - let me know if you want to discuss further.

src/protocols/displayport.stanza
bhusang: I don't recall that we did. If there is an option for fewer number of lanes, I can make this an argument. AFAIK, miniDP is still 4 lanes.

src/protocols/displayport.stanza
bhusang: I remember seeing this on ICs so I kept it here. I think that it might be used by the interface for sync?

src/bundles/comms.stanza
bhusang: Yeah, let's chat about this.

src/protocols/displayport.stanza
callendorph: No - then nevermind.

src/bundles/comms.stanza
bhusang: Keeping COPI / CIPO

src/bundles/comms.stanza
bhusang: Removing extra pins, would love to find a way to do bundle inheritance as well to extend the base bundle for vendor-specific extensions.

PR #66, opened 2024-06-20
remove validate! function

lchao-jitx:
remove validate! function


callendorph:
For Documentation Purposes: 

We've removed this because the `validate!` function in JSL is not as good as the validation already exists in the JITX runtime. It was too restrictive compared to what JSL actually supports.

The idea is that the user should just rely on the JITX runtimes reporting of errors and if that is not sufficient we should update the JITX runtime error reporting.

callendorph (approved):
👍  - Thanks @lchao-jitx

PR #71, opened 2024-06-25
JSL/make-layer-stack: explore different methods to create LayerStack

lchao-jitx:
I tried to explore different methods to create LayerStack.
Check out 
- deftest(layerstack) make-layer-stack-method1 :
- deftest(layerstack) make-layer-stack-method2 :
- deftest(layerstack) make-layer-stack-method3 :
- deftest(layerstack) make-layer-stack-method4 :

### DIFF
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -574,7 +574,7 @@ doc:\<DOC>
        soldermask
      ]
   @param name The name of the LayerStack
-  @param top-layers The pairs of mechanical and dielectric layaers
+  @param top-layers The pairs of mechanical and dielectric layers
   @param description The description of the LayerStack
 <DOC>
 public defn make-layer-stack (name:String, outers:Tuple<[LayerSpec, LayerSpec]>
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -577,7 +577,7 @@ doc:\<DOC>
        soldermask
      ]
   @param name The name of the LayerStack
-  @param top-layers The pairs of mechanical and dielectric layers
+  @param top-layers The pairs of mechanical and dielectric layers, from outer to inner layers.
   @param description The description of the LayerStack
 <DOC>
 public defn make-layer-stack (name:String, outers:Tuple<[LayerSpec, LayerSpec]>
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -567,10 +567,13 @@ doc:\<DOC>
   ```
   The LayerStack created has these layers:
      [ soldermask
-       copper-35um prepreg-2313
-       copper-17_5um core-2313
-       copper-17_5um core-2313
-       copper-35um prepreg-2313
+       copper-35um
+       prepreg-2313
+       copper-17_5um
+       core-2313
+       copper-17_5um
+       prepreg-2313
+       copper-35um
        soldermask
      ]
   @param name The name of the LayerStack
@@ -583,7 +586,7 @@ public defn make-layer-stack (name:String, outers:Tuple<[LayerSpec, LayerSpec]>
   ;Access pairs of outer-layers in reverse order
   for outer in in-reverse(outers) do:
     val [metal, insul] = outer
-      add-symmetric(metal, insul, stack)
+    add-symmetric(metal, insul, stack)
   if value?(soldermask) is LayerSpec :
     add-soldermask(value!(soldermask), stack) 
   stack
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -549,35 +549,27 @@ public defn create-pcb-stackup (stack:LayerStack) :
   custom-stackup
 
 doc:\<DOC>
-  Construct a LayerStack from a tuple of top layers and create the bottom symmetric layers. 
+  Construct a LayerStack with a tuple of outer layers. 
 
   Example:
   ```
-    val jlcpcb-jlc2313 = make-layer-stack("JLCPCB 4-layer 1.6mm",
-      [
-        soldermask
-        copper-35um 
-        prepreg-2313 
-        copper-17_5um
-        core
-      ]) where :
-        val soldermask = Soldermask(0.019, Soldermask-Material-BSN4000)
-        val copper-35um = Copper(0.035)
-        val prepreg-2313 = FR4(0.1, FR4-Material-2313)
-        val copper-17_5um = Copper(0.0175)
-        val core = FR4(1.265, FR4-Material-Core)
+    val jlcpcb-jlc2313 = make-layer-stack("JLCPCB 4-layer 1.6mm", outer-layers) where :
+      val copper-35um = Copper(0.035)
+      val prepreg-2313 = FR4(0.1, FR4-Material-2313)
+      val copper-17_5um = Copper(0.0175)
+      val core = FR4(1.265, FR4-Material-Core)
+      val outer-layers = [
+        [copper-35um prepreg-2313] 
+        [copper-17_5um core]
+      ]
   ```
 <DOC>
-public defn make-layer-stack (name:String, top-layers:Tuple<LayerSpec>) -> LayerStack :
+public defn make-layer-stack (name:String, outers:Tuple<[LayerSpec, LayerSpec]>) -> LayerStack :
   ;Access top-layers with reversed index
   ; - so the middle layer is index 0
-  defn reversed-layers (idx:Int) :
-    top-layers[length(top-layers) - 1 - idx]
   val stack = LayerStack(name = name)
-  add-symmetric(reversed-layers(1), reversed-layers(0), stack)
-  for idx in 2 to length(top-layers) by 2 do :
-    if is-soldermask(reversed-layers(idx)) :
-      add-soldermask(reversed-layers(idx), stack)
-    else :
-      add-symmetric(reversed-layers(idx + 1), reversed-layers(idx), stack)
+  for outer in in-reverse(outers) do:
+    val [metal, insul] = outer
+      add-symmetric(metal, insul, stack)
   stack
+
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -314,7 +314,7 @@ public defstruct LayerStack :
 
   layers:Vector<LayerSpec> with: (
     updater => sub-layers
-  )
+    default => Vector<LayerSpec>())
 with:
   constructor => #LayerStack
   printer => true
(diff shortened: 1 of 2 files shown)


tests/layerstack.stanza
```diff
@@ -196,5 +196,106 @@ deftest(layerstack) test-copper-optional-arguments :
...
+    val soldermask = Soldermask(0.019, SoldermaskMaterial)
+    val copper-35um = Copper(0.035)
+    val prepreg-2313 = FR4(0.1, FR4-Material)
+    val copper-17_5um = Copper(0.0175)
+    val core = FR4(1.265, FR4-Material)
+    add-soldermask(soldermask,
+      add-symmetric(copper-35um, prepreg-2313,
+        add-symmetric(copper-17_5um, core, stack)
+      )
+    )
```
callendorph, line 259: I don't see a strong reason to deviate from this pattern. 

The `make-layer-stack` looks complicated. We would need to write a LOT of tests to make sure this works correctly in all of the corner cases. 

Current API (ie, `add-symmetric`, `add-top`, etc) is very straightforward.

Why would we not use this form ? 

@emspin  - thoughts ?

emspin: Sorry just getting around to this. 

Thanks @lchao-jitx for coding this out so that we can see it in full form. My evaluation is that overall `make-layer-stack` doesn't do enough to simplify the problem for it to be worth it. 

The general use-case is how many layers @callendorph? I'm going to guess it's usually not past 10ish right? 

I think at that scale, having the nested `add-*` functions is still fairly grokable and handleable. If we were dealing with 10+ layers then I could argue in favor of having the structure of `make-layer-stack`, even if only from a coding aesthetic/manageability standpoint. 

As Carl mentioned there is another notable cost in testing all the edge cases, which again I would feel justified in doing if we were gaining more from the implementation, but for now I suggest we do not spend time there. 

@lchao-jitx does that sound like a reasonable perspective?

callendorph: >The general use-case is how many layers @callendorph? I'm going to guess it's usually not past 10ish right?

Sub-10 is probably where the mean is. There is a fat tail for larger configurations though.

It doesn't have to be nested in the general case. :

```
val stack = LayerStack()
...
add-symmetric(copper-35um, core, stack).                ; Inner Most Core
add-symmetric(copper-35um, prepreg-2313, stack)
add-symmetric(copper-35um, core, stack)
add-symmetric(copper-35um, prepreg-2313, stack) ; Top copper
```

The only issue is that now is is reverse ordered, which is unfortunate. I think that is both less readable and if reach the tipping point - you might have to do a lot of error prone work to put it in the non-nested form.

Another possible way to address large stackups would be:

```
val stack = LayerStack()
...
val outers = [
   [copper-35um, prepreg-2313]
   [copper-35um, core]
   [copper-35um, prepreg-2313]
   [copper-35um, core]
]
for  outer in in-reverse(outers) do:
  val [metal, insul] = outer
   add-symmetric(metal, insul, stack)

```
 
This reuses the existing `add-symmetric` and allows for the listing of the layers in a tuple.

lchao-jitx: I revised the `make-layer-stack` function to use the above method to provide the outer layers in a tuple of pairs of mechanical and dielectric layers.

It looks like this:
```
doc:\<DOC>
  Construct a LayerStack with a tuple of outer layers. 

  Example:
  ```
    val jlcpcb-jlc2313 = make-layer-stack("JLCPCB 4-layer 1.6mm", outer-layers) where :
      val copper-35um = Copper(0.035)
      val prepreg-2313 = FR4(0.1, FR4-Material-2313)
      val copper-17_5um = Copper(0.0175)
      val core = FR4(1.265, FR4-Material-Core)
      val outer-layers = [
        [copper-35um prepreg-2313] 
        [copper-17_5um core]
      ]
  ```
<DOC>
public defn make-layer-stack (name:String, outers:Tuple<[LayerSpec, LayerSpec]>) -> LayerStack :
  ;Access top-layers with reversed index
  ; - so the middle layer is index 0
  val stack = LayerStack(name = name)
  for outer in in-reverse(outers) do:
    val [metal, insul] = outer
      add-symmetric(metal, insul, stack)
  stack
```

In `tests/layerstack.stanza`, there are still 4 methods to create the layer stack.
  - Method 1:  Use a series of add-symmetric functions to create a LayerStack
      - 1A: use `let` and `val`'s
      - 1B: put them in a function
  - Method 2: Use the function make-layer-stack to create a LayerStack from a tuple of outer layers
     - 2A: use `let` and `val`'s
     - 2B: put them in a function
     
Questions:
  - Should I keep these in the `jsl/tests/layerstack` as examples?
  - Which one should I use for the JLCPCB repo?

callendorph: Let's proceed with `2A` for the JLCPCB repo.
Let's keep only the `2A` version as an example in `jsl/tests/layerstack`

Encapsulating these lines:

```  
  val copper-35um = Copper(0.035)
  val prepreg-2313 = FR4(0.1, FR4-Material-2313)
  val copper-17_5um = Copper(0.0175)
  val core = FR4(1.265, FR4-Material-Core)
```

In the where clause is the probably not the right approach in the general case.

The `copper` statements in particular are likely going to be reused in many stackups.  The `prepreg` and `core` will likely also be reused but not in as many. By putting the in the `where` clause - they can't be reused. 

I think this approach works well for the `outer-layers` list of tuples.

The `make-layer-stack` approach as written works if we can:

1.   Let's add a `description` argument so that the user can set that if they so desire. 
2.  Let's add an keyword optional `soldermask` argument


That way we can do something like:

```
val sm = Soldermask(0.019, SoldermaskMaterial)
val stack = make-layer-stack("Name", outers, soldermask = sm) where:
   ...

```

If the user doesn't provide a `soldermask` arg - then none is applied.

src/layerstack.stanza
```diff
@@ -536,3 +547,29 @@ public defn create-pcb-stackup (stack:LayerStack) :
...
+  ```
+    val jlcpcb-jlc2313 = make-layer-stack("JLCPCB 4-layer 1.6mm", outer-layers) where :
+      val copper-35um = Copper(0.035)
+      val prepreg-2313 = FR4(0.1, FR4-Material-2313)
+      val copper-17_5um = Copper(0.0175)
+      val core = FR4(1.265, FR4-Material-Core)
+      val outer-layers = [
+        [copper-35um prepreg-2313] 
+        [copper-17_5um core]
+      ]
```
callendorph, line 564: Let's please add to the documentation here a description of what the stackup is expected to be after this function runs.

bhusang:
I would like to do a webinar on setting up a layer stack for a random PCB process. The examples I would like to cover are a standard 1 layer flex, 2 layer rigid, 4 layer rigid, 8 layer rigid (with 2oz copper), and 20 layer rigid. Can this PR be helpful to that end? If so, can I get some pointers to the code that would generate the above examples? @lchao-jitx @callendorph @emspin

callendorph:
Yes - I think using the `make-layer-stack` that @lchao-jitx has put together would be a good idea. 

I believe @lchao-jitx  still has outstanding changes to make on this implementation before it merges.

tests/layerstack.stanza
lchao-jitx: I have added an optional parameter for `soldermask` for `make-layer-stack` function.

I see that copper-35um and copper-17_7um are likely common. But the thickness of the prereg and core differ between different jlcpcb stackups. So, initially I tried to share them between different jlcpcb stackups. But then later I put all of them inside `where`.

 Here is one example remains for jlcpcb in `tests`:
 ```
   val stack = make-layer-stack("JLCPCB 4-layer 1.6mm", top-layers, soldermask = soldermask) where :
    val soldermask = Soldermask(0.019, SoldermaskMaterial)
    val copper-35um = Copper(0.035)
    val prepreg-2313 = FR4(0.1, FR4-Material)
    val copper-17_5um = Copper(0.0175)
    val core-2313 = FR4(1.265, FR4-Material)
    val top-layers =  [
      [copper-35um prepreg-2313] 
      [copper-17_5um core-2313]
    ]
 ```

lchao-jitx: For the jlcpcb repo, I see that we will have one stanza file for each stackup. So, we won't be sharing the common `val`'s, right? 
Unless we put them in one common stanza file to share?

> In the where clause is the probably not the right approach in the general case.

We have these options if we don't put them all inside `where`:
1. put `val` for `copper-35um`, `prereg-2313` inside `where` like the example above
2. put `val` for `copper-35um`, `prereg-2313` outside:
```
 val soldermask = Soldermask(0.019, SoldermaskMaterial)
 val copper-35um = Copper(0.035)
 val prepreg-2313 = FR4(0.1, FR4-Material)
 val copper-17_5um = Copper(0.0175)
 val core-2313 = FR4(1.265, FR4-Material)
 val stack = make-layer-stack("JLCPCB 4-layer 1.6mm", top-layers, soldermask = soldermask) where :
   val top-layers =  [
     [copper-35um prepreg-2313] 
     [copper-17_5um core-2313]
   ]
```
3. Put the common ones outside of where.
```
  val copper-35um = Copper(0.035)
  val copper-17_5um = Copper(0.0175)
  val stack = make-layer-stack("JLCPCB 4-layer 1.6mm", top-layers, soldermask = soldermask) where :
    val soldermask = Soldermask(0.019, SoldermaskMaterial)
    val prepreg-2313 = FR4(0.1, FR4-Material)
    val core-2313 = FR4(1.265, FR4-Material)
    val top-layers =  [
      [copper-35um prepreg-2313] 
      [copper-17_5um core-2313]
    ]
```
Which one would you prefer?

src/layerstack.stanza
lchao-jitx: Added it:
```
doc:\<DOC>
  Construct a LayerStack with a tuple of outer layers. 

  Example:
  ```
    val copper-35um = Copper(0.035)
    val copper-17_5um = Copper(0.0175)
    val jlcpcb-jlc2313 = make-layer-stack("JLCPCB 4-layer 1.6mm", outer-layers,
      soldermask = soldermask) where :
      val soldermask = Soldermask(0.019, SoldermaskMaterial)
      val prepreg-2313 = FR4(0.1, FR4-Material-2313)
      val core-2313 = FR4(1.265, FR4-Material-Core)
      val outer-layers = [
        [copper-35um prepreg-2313] 
        [copper-17_5um core-2313]
      ]
  ```
  The LayerStack created has these layers:
     [ soldermask
       copper-35um prepreg-2313
       copper-17_5um core-2313
       copper-17_5um core-2313
       copper-35um prepreg-2313
       soldermask
     ]
  @param name The name of the LayerStack
  @param top-layers The pairs of mechanical and dielectric layers
  @param description The description of the LayerStack
<DOC>
```

tests/layerstack.stanza
lchao-jitx: One more question:
I have added an optional `description` parameter to `make-layer-stack`. The `name` parameter is still required. I see that for the other functions, both `name` and `description` are optional parameters. Should I make both `name` and `description` optional parameters to be consistent with the other functions?

src/layerstack.stanza
```diff
@@ -536,3 +547,44 @@ public defn create-pcb-stackup (stack:LayerStack) :
...
+        [copper-35um prepreg-2313] 
+        [copper-17_5um core-2313]
+      ]
+  ```
+  The LayerStack created has these layers:
+     [ soldermask
+       copper-35um prepreg-2313
+       copper-17_5um core-2313
+       copper-17_5um core-2313
+       copper-35um prepreg-2313
```
callendorph, line 573: This is incorrect - the resultant layer structure should be:

```
[
  soldermask
  copper-35um 
  prepreg-2313
  copper-17_5um 
  core-2313
  copper-17_5um 
  prepreg-2313
  copper-35um 
  soldermask
]
```

Notice the singular `core-2313` in the center.

lchao-jitx: I noticed that and already corrected it in the last commit.

src/layerstack.stanza
lchao-jitx: Edited the comment above with the corrected layers.

tests/layerstack.stanza
lchao-jitx: In the jlc-pcb repo, I have created the stack like this so all of the `val` for layers are outside of `where`:
```
defpackage jlc-pcb/JLC04161H-7628:
  import core
  import jitx
  import jsl/layerstack
  
val copper-35mm = Copper(0.035, name = "cu")
val copper-75mm = Copper(0.070, name = "cu75") 
val core-FR4 = FR4(1.065, name = "core")
val prepreg-FR4 = FR4(0.2104, name = "7628x1")
public val stack = make-layer-stack("JLC04161H-7628", outers) where :
  val outers = 
    [ 
      [copper-75mm prepreg-FR4]
      [copper-35mm core-FR4]
    ]

public val stackup = create-pcb-stackup(stack)

```
I have modified the deftest in `jsl/tests/layerstack` to use the same structure.

callendorph (approved):
:+1: Thank you for all the hard work on this implementaiton!

PR #73, opened 2024-07-08
Added initial Via Structures Framework

callendorph:
This PR adds a new framework for constructing via structures using `pcb-module` with `geom/via/layer` statements. 

Here are some examples: 

Single Ended Via Structure with a Double Chipped Circle Antipad and 4 ground cage vias: 
 
Screenshot 2024-07-07 at 9 58 14 PM

Diff-Pair Via Structure with a Double Notched Rectangle Antipad and 6 ground cage vias: 
Screenshot 2024-07-07 at 9 58 50 PM

Hard to see the double notch anti-pad - here is the keepout region turned on: 
Screenshot 2024-07-07 at 9 58 55 PM

# Issues:

1.  The `CustomLayer` content was not showing in the layer stack - so I put the insertion points on the silkscreen just so they would show. Probably want to revert that in the future.
2. I can't create a courtyard by introspection. There isn't really a way for me to introspect the via locations in a module. Or at least, I haven't found one. If you know of a way - let me know.
3.  The forbid copper regions seem to have an additional clearance between the shape and the copper on the internal layer.  
4.  Pin Assignment based usage (ie, with `supports` statements) is not working. I think this may be running into an ambiguous topology case because of the vias.

### DIFF
--- a/examples/via-structures/diff-via-structure.stanza
+++ b/examples/via-structures/diff-via-structure.stanza
@@ -1,0 +1,64 @@
+#use-added-syntax(jitx)
+defpackage jsl/examples/via-structures/diff-via-structures:
+  import core
+  import jitx
+  import jitx/commands
+
+  import jsl/via-structures
+  import jsl/si/helpers
+  import jsl/si/pairs
+  import jsl/si/couplers
+
+  import jsl/examples/protocols/common/example-board
+  import jsl/examples/protocols/common/example-components
+
+
+pcb-module top-level :
+  val diff-via = DifferentialViaStructure(
+    name = "DF-1",
+    via-def = uTop-2,
+    pitch = 0.4
+  )
+
+  val gc = PolarViaGroundCage(
+    via-def = default-TH,
+    count = 12,
+    radius = 0.6,
+    skips = [2, 3, 4, 8, 9, 10]
+  )
+  add-ground-cage(diff-via, gc)
+  add-anti-pad(
+    diff-via,
+    SimpleAntiPad(
+      shape = RoundedRectangle(0.8, 0.4, 0.1),
+      start = LayerIndex(0, Top),
+      end = LayerIndex(1,Top))
+  )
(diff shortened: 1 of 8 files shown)


src/via-structures.stanza
```diff
@@ -0,0 +1,617 @@
...
+defmethod make-ground-cage (v:PolarViaGroundCage, n:JITXObject -- pose:Pose) -> False:
+
+  val valid-pos = for i in 0 to count(v) filter:
+    not contains?(skips(v), i)
+
+  inside pcb-module:
+    geom(n):
+      for i in valid-pos do:
+        val pos = pose * compute-pos(v, i)
+        via(via-def(v)) at apply(pos, Point(0.0, 0.0))
```
bhusang, line 149: My my info:
is this the same as 
`at Point(0.0, 0.0) * pos`
?

callendorph: Ya - I think this was just a brain fart. I tried using a `Pose` after the `at` and that didn't work - it was looking for a point. I could have done `at pos * Point(0.0, 0.0)`

examples/via-structures/diff-via-structure.stanza
```diff
@@ -0,0 +1,65 @@
...
+    count = 12,
+    radius = 0.6,
+    skips = [2, 3, 4, 8, 9, 10]
+  )
+  add-ground-cage(diff-via, gc)
+  add-anti-pad(
+    diff-via,
+    SimpleAntiPad(
+      shape = DoubleNotchedRectangle(width = 0.8, height = 0.4, notch-width = 0.4, notch-height = 0.15)
+      ; shape = RoundedRectangle(0.8, 0.4, 0.1),
```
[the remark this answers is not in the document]
callendorph: Ya - I was trying to just indicate that there are multiple options here but wasn't explicit.

bhusang (approved):

PR #74, opened 2024-07-09
Corner cut BGA

alijitx:
Adds a corner cut matrix planner for the BGA landpattern generation. Fixes a bug where the BGA silkscreen generation would crash if no A[1] pad exists by selecting the lowest-numbered alternative pad as the substitute.

### DIFF
--- a/src/landpatterns/VirtualLP.stanza
+++ b/src/landpatterns/VirtualLP.stanza
@@ -531,7 +531,9 @@ public defmethod make-landpattern (vp:VirtualLP, parent-pose:Pose = DEF_VLP_POSE
     for child in children(vp) do:
       make-landpattern(child, curr-pose)
 
-
+doc: \<DOC>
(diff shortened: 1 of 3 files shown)


src/landpatterns/VirtualLP.stanza
```diff
@@ -564,8 +564,54 @@ public defn get-pad-by-ref! (vp:VirtualLP, r:Ref|Int) -> VirtualPad :
     (_:None): throw $ ValueError("Failed to find Pad with Ref: %_" % [r])
     (x:One<VirtualPad>): value(x)
 
+public defn get-first-pad (vp:VirtualLP) -> VirtualPad :
```
callendorph, line 567: Can we please add docstrings for the public functions

callendorph (approved):
Looks good. 
Let's add docstrings for pad id comparison functions.

PR #75, opened 2024-07-09
Bundle pin rename

alijitx:
Renames the member pin names for the general bundles to distinct names instead of 'p' to ensure the bundles cannot be aliased to each other.

### DIFF
--- a/src/bundles/general.stanza
+++ b/src/bundles/general.stanza
@@ -100,39 +100,55 @@ public pcb-bundle power :
 doc: \<DOC>
 GPIO Interface Bundle
 
-@member p GPIO Pin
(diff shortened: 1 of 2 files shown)


bhusang (approved):

callendorph:
Have you run the unit tests locally for this branch ?

bhusang:
I assume that you are asking @alijitx ?

callendorph:
Yes - I'm just trying to head off a regression. I thought that there was a bundles unit test that ref'd these types.

bhusang (approved):

alijitx:
I ran the tests, the only ones that failed seemed to be the same ones from PROD-272

PR #77, opened 2024-07-11
Add a configuration strap component array generator

callendorph:
This adds a new module that will array a set of components for a configuration strap array.

https://github.com/JITx-Inc/jsl/assets/622392/7a92b104-c964-407b-8996-98d6068eeddc

It works for through-hole and surface mounts and handles figuring out how to best orientate the components when placing them. 

Then you can use pin assignment to generate the best connections. 

Screenshot 2024-07-10 at 9 45 27 PM

### DIFF
--- a/src/design/ConfigStrap.stanza
+++ b/src/design/ConfigStrap.stanza
@@ -1,0 +1,224 @@
+doc: \<DOC>
+@brief Configuration Strap Module
+
+This package contains a module generator that can be
+used to define a series of configuration straps.
+
+<DOC>
+#use-added-syntax(jitx)
+defpackage jsl/design/ConfigStrap:
+
+  import core
+  import collections
+  import jitx
+  import jitx/commands
+
+  import jsl/errors
+  import jsl/ensure
+  import jsl/design/introspection
+  import jsl/landpatterns/framework
+
+public deftype ArrayPacking
+doc: \<DOC>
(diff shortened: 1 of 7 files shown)


src/design/ConfigStrap.stanza
```diff
@@ -0,0 +1,224 @@
...
+
+
+doc: \<DOC>
+Configuration Bus Bundle
+
+This bundle is what is used to access the
+pin assignment supports for the ConfigStrap module.
+<DOC>
+public pcb-bundle config-strap-bus (num-elems:Int) :
+  port config : pin[num-elems]
```
bhusang, line 126: You're coding a bundle in the designs folder. No real issue except for bookkeeping (bundles versus design). Ok for you?

bhusang: The bundle is rather design-specific though. So it makes sense.

callendorph: Yes - This bundle is only applicable in this application. We do a similar thing with the bundles in the protocols directory where a bundle is only applicable to that application. 

I put it in `design` because it didn't seem to fit anywhere else. We could maybe make a `generators` or `circuits` directory or something ?

callendorph: I moved it to `circuits` but kept the structure.

bhusang (approved):

PR #79, opened 2024-07-12
Jsl/make layer stack v2

lchao-jitx:
- Revert the last commit for 'Enhance make-layer-stack to handle multiple adjacent prepreg layers'
and then re-commit it for review. 
   - This commit also removes the reference to JLCPCB in the examples/tests for layer-stack so that a generic 4-layer stack is used.
- Revise `verify-layers` to handle even number of layers in a stack.

In summary, the make-layer-stack is revised to support these two examples:
```
    make-layer-stack("4-layer stack with three adjacent prepreg layers", top-layers) where :
      val top-layers =  [
        [copper-35um [prepreg prepreg2 prepreg3]]
        [copper-17_5um core]
      ]
```
and
```
    make-layer-stack("6-layer stack with even number of layers", outers, even-layers? = true) where :
      val outers =
        [
          [copper-35um prepreg]
          [copper-17_5um core]
          [copper-17_5um prepreg2] ; => two layers of prepreg2 in the center
        ]
```

### DIFF
--- a/src/layerstack.stanza
+++ b/src/layerstack.stanza
@@ -355,9 +355,11 @@ When used on its own, this function is useful for creating asymmetric
 pcb stackups. This function is most useful when used on a non-empty
 layer stack.
 
-@param l-set layer or set of layers to add to the top of the stack. Note
+@param l-set layer or set of layers to add to the top of the stack, ordered from inner to outer (bottom to top). Note
 that if this value is a collection, then the layers will be added to
 the stack one at a time. For example:
+
+
 
 ```
 add-top( [dielectric, copper], stack)
@@ -389,7 +391,7 @@ When used on its own, this function is useful for creating asymmetric
 pcb stackups. This function is most useful when used on a non-empty
 layer stack.
 
-@param l-set layer or set of layers to add to the top of the stack. Note
+@param l-set layer or set of layers to add to the top of the stack, ordered from inner to outer (top to bottom). Note
 that if this value is a collection, then the layers will be added to
 the stack one at a time. For example:
 
@@ -447,9 +449,9 @@ add-symmetric(copper-35um, core-FR4,
 <DOC>
 public defn add-symmetric (copper:LayerSpec, dielectric:LayerSpec, s:LayerStack) -> LayerStack:
   if length(layers(s)) > 0:
-    val ls = [dielectric, copper]
-    add-bottom(ls, s)
+    val ls = [dielectric copper]
     add-top(ls, s)
+    add-bottom(ls, s)
   else:
     ; If there are no layers yet - then we need to just create
     ;  a single dielectric layer.
@@ -555,11 +557,11 @@ doc:\<DOC>
   ```
     val copper-35um = Copper(0.035)
     val copper-17_5um = Copper(0.0175)
+    val core = FR4(1.265, FR4-Material)
     val stack = make-layer-stack("4-Layer Stack with Soldermask", outer-layers,
       soldermask = soldermask) where :
       val soldermask = Soldermask(0.019, SoldermaskMaterial)
       val prepreg = FR4(0.1, FR4-Material)
-      val core = FR4(1.265, FR4-Material)
       val outer-layers = [
         [copper-35um prepreg] 
         [copper-17_5um core]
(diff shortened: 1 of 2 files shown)


lchao-jitx:
On second thought, maybe a better method to support multiple adjacent prepreg layers is:
```
    val stack = make-layer-stack("4-layer with two adjacent prepreg layers", top-layers) where :
      val top-layers =  [
        [copper-35um prepreg prepreg]
        [copper-17_5um core]
      ]
```
rather than what I have in the last commit: 'Enhance make-layer-stack ...'
```
    val stack = make-layer-stack("4-layer with two adjacent prepreg layers", top-layers) where :
      val top-layers =  [
        [copper-35um prepreg]
        prepreg
        [copper-17_5um core]
      ]
```

callendorph:
I would prefer that we consider the following: 

```
    val stack = make-layer-stack("4-layer with two adjacent prepreg layers", top-layers) where :
      val top-layers =  [
        [copper-35um [prepreg prepreg] ]
        [copper-17_5um core]
      ]
```

Because they we can add a function like:

```
defn make-multi-ply (m:Material, num-ply:Int ) -> Tuple<Material> : 
  to-tuple $ for i in 0 to num-ply seq:
      m

defn make-double-ply (m:Material) -> Tuple<Material> : 
  make-multi-ply(m, 2)
```

And this ends up looking more like: 

```
    val prepreg-x2 = make-dual-ply(prepreg)
    val stack = make-layer-stack("4-layer with two adjacent prepreg layers", top-layers) where :
      val top-layers =  [
        [copper-35um prepreg-x2 ]
        [copper-17_5um core]
      ]
```

This was we can prevent the introduction of touching copper layers by design and we can support multiple pre-preg layers as necessary.

lchao-jitx:
I found that some 6-layer JLCPCB stakcups have an even number of layers, such as "JLC061211-2116", "JLC061211-2116" , "JLC06201H-3313A", "JLC062011-7628", "JLC062011-3313", "JLC062011-1080"  etc.
The version in the last commit works for those cases since whenever it is a single LayerSpec, it is added to both the top and the bottom of the stack.
image
For example,
```
  val test-stack = make-layer-stack("JLC06201H-3313A", outers) where :
  val core = FR4("core", 0.7, "core")
  val prepreg = FR4("2116", 0.1164, "2116*1")
  val prepregA = FR4("3313", 0.0994, "3313*1")
  val outers =
    [
      [copper-35um prepreg]
      [copper-15_2um core]
      copper-15_2um
      prepregA       ; => two layers of prepregA in the center
    ]
```

lchao-jitx:
The two adjacent prepreg layers may not be the same though.
Here is an example:
image
We could support it this way.
```
protected val stack = make-layer-stack("JLC04161H-2116A", outers) where :
  val core = FR4("core", 0.6, "core")
  val prepreg = FR4("2116", 0.124, "2116*1")
  val prepregA = FR4("7628", 0.218, "7628*1")
  val prepregB = FR4("2116", 0.1164, "2116*1")
  val outers = 
    [ 
      [copper-35um [prepreg prepregA prepregB]]
      [copper-15_2um core]
    ]
  ```

**Comment by lchao-jitx** [comment]
For the case when there are even number of layers (no one center layer), perhaps we can support it this way:
```
val test-stack = make-layer-stack("JLC06201H-3313A", outers) where :
  val core = FR4("core", 0.7, "core")
  val prepreg = FR4("2116", 0.1164, "2116*1")
  val prepregA = FR4("3313", 0.0994, "3313*1")
  val outers =
    [
      [copper-35um prepreg]
      [copper-15_2um core]
      [copper-15_2um [prepregA]]       ; => two layers of prepregA in the center
    ]
```
To use `[copper-15_2um [prepregA]]`, rather than `[copper-15_2um prepregA]`.

But then, maybe it is better to just use an optional flag `even-layers`, like I used for `verify-layers` function. 
This way it's more explicit for the user.

**Comment by lchao-jitx** [comment]
The `make-layer-stack` is revised to support these two examples:
```
    make-layer-stack("4-layer stack with three adjacent prepreg layers", top-layers) where :
      val top-layers =  [
        [copper-35um [prepreg prepreg2 prepreg3]]
        [copper-17_5um core]
      ]
```
and 
```
    make-layer-stack("6-layer stack with even number of layers", outers, even-layers? = true) where :
      val outers =
        [
          [copper-35um prepreg]
          [copper-17_5um core]
          [copper-17_5um prepreg2] ; => two layers of prepreg2 in the center
        ]
```

File: src/layerstack.stanza [hunk 1]
```diff
@@ -355,10 +355,12 @@ When used on its own, this function is useful for creating asymmetric
 pcb stackups. This function is most useful when used on a non-empty
 layer stack.
 
-@param l-set layer or set of layers to add to the top of the stack. Note
+@param l-set layer or set of layers to add to the top of the stack, ordered from inner to outer (bottom to top). Note
```
callendorph, line 358: Let's please remove `(bottom to top)` - this is confusing because we also talk about the bottom side of the board vs the top side of the board. 

I would be OK with `ordered from inner-most to outer -most on the top side`

src/layerstack.stanza
```diff
@@ -389,7 +391,7 @@ When used on its own, this function is useful for creating asymmetric
 pcb stackups. This function is most useful when used on a non-empty
 layer stack.
 
-@param l-set layer or set of layers to add to the top of the stack. Note
+@param l-set layer or set of layers to add to the top of the stack, ordered from inner to outer (top to bottom). Note
```
callendorph, line 394: Same comment here - I think this is confusing. Let's re-word this as: 

`ordered from inner-most to outer-most on the bottom side`

src/layerstack.stanza
```diff
@@ -447,9 +449,9 @@ add-symmetric(copper-35um, core-FR4,
 <DOC>
 public defn add-symmetric (copper:LayerSpec, dielectric:LayerSpec, s:LayerStack) -> LayerStack:
   if length(layers(s)) > 0:
-    val ls = [dielectric, copper]
-    add-bottom(ls, s)
+    val ls = [dielectric copper]
     add-top(ls, s)
+    add-bottom(ls, s)
```
callendorph, line 454: Why did we do this ?

src/layerstack.stanza
```diff
@@ -576,49 +579,106 @@ doc:\<DOC>
        copper-35um
        soldermask
      ]
+    ```
   
-  Example 2: 4-layer with two adjacent prepreg layers
+  Example 2: 4-layer stack with three adjacent prepreg layers, such as "JLC04161H-7628B"
```
callendorph, line 584: Please remove the reference to JLC-PCB here.

src/layerstack.stanza
```diff
@@ -576,49 +579,106 @@ doc:\<DOC>
...
+       core
+       copper-17_5um
+       prepreg3
+       prepreg2
+       prepreg
+       copper-35um
+     ]
+    ```
+
+  Example 3: 6-layer stack with even number of layers, such as "JLC06201H-3313A"
```
callendorph, line 610: For examples 1, 2, 3 - let's please use the `@snippet` marker and related `@snip-note` marker syntax here for the docstrings.

See here for more details:
https://jit-x.slack.com/docs/TKJEB34LS/F05RW135Y0J

src/layerstack.stanza
```diff
@@ -576,49 +579,106 @@ doc:\<DOC>
...
+  ```
+    make-layer-stack("6-layer stack with even number of layers", outers, even-layers? = true) where :
```
callendorph, line 612: Why would we do this ? 
Why would we not just continue using the pattern we have already established?

For example: 

```
     val outers =
        [
          [copper-35um prepreg]
          [copper-17_5um core]
          [copper-17_5um [prepreg2, prepreg2]] ; => two layers of prepreg2 in the center
        ]

```

I would much prefer this than `even-layers?` which is difficult to explain and requires additional handling.

src/layerstack.stanza
```diff
@@ -576,49 +579,106 @@ doc:\<DOC>
...
+        if n-insuls == 1 : ; [copper [dielectric]] => treat it as [copper dielectric]
+          add-symmetric(metal, insuls[0], stack)
+        else :
+          add-symmetric(insuls[n-insuls - 2], insuls[n-insuls - 1], stack)
+          ; then add the rest of the layers
+          val layers = to-list $ cat([metal], insuls[0 to n-insuls - 2]) ; from outer to inner layers
+          add-top(reverse(layers), stack) ; ordered from inner to outer
+          add-bottom(reverse(layers), stack)
+    else : ; even-layers? = true
+      if outer[1] is LayerSpec :
```
callendorph, line 672: Can we please use a `match` statement instead of using `if ... is` ? 

The `if ... is` is an anti-pattern in stanza and the idiomatic mechanism is the `match` clause.

src/layerstack.stanza
```diff
@@ -576,49 +579,106 @@ doc:\<DOC>
...
+        val [metal insul] = outer as [LayerSpec LayerSpec]
         add-symmetric(metal, insul, stack)
+      else : ; outer[1] is Tuple<LayerSpec>
+        val [metal insuls] = outer as [LayerSpec Tuple<LayerSpec>]
+        ; add the innermost layers first
+        val n-insuls = length(insuls) ; insuls[n-insuls - 1] is the innermost layer
+        if n-insuls == 1 : ; [copper [dielectric]] => treat it as [copper dielectric]
+          add-symmetric(metal, insuls[0], stack)
+        else :
+          add-symmetric(insuls[n-insuls - 2], insuls[n-insuls - 1], stack)
```
callendorph, line 666: This doesn't look right. What happens if  `n-insuls` is greater than 3 ?

src/layerstack.stanza
```diff
@@ -576,49 +579,106 @@ doc:\<DOC>
...
-      (outer:LayerSpec):
-        ; add one prepreg layer to top and bottom
-        add-top(outer, stack)
-        add-bottom(outer, stack)
-      ([metal, insul]:[LayerSpec LayerSpec]):
-        ; add [copper, prepreg] layers to top and bottom
+    ;Verify input
+    if outer[1] is Tuple<LayerSpec> and empty?(outer[1] as Tuple<LayerSpec>) :
+      throw(Exception("Invalid argument: The tuple of dialectric layers is empty. Expect [LayerSpec Tuple<LayerSpec>]. Found %_" % [outer]))
+    if not even-layers? :
```
callendorph, line 655: The `even-layers?` parameter here is making this function about 1000x more complicated than this needs to be.

callendorph: We need to remove this.

lchao-jitx:
Please see the third version of `make-layer-stack` in [https://github.com/JITx-Inc/jsl/pull/92](https://github.com/JITx-Inc/jsl/pull/92) with a much simplified interface and logic.

I have incorporated changes to address the comments in this PR there too.

lchao-jitx:
Dropping this PR since the third version of `make-layer-stack` in [this other PR ](https://github.com/JITx-Inc/jsl/pull/92)has a much cleaner interface.

PR #82, opened 2024-07-15
Removed ref to JITX package removed in latest `jitx-client` build

callendorph:
This package has been removed in the latest `jitx-client` on `develop`. It causes JSL to break. I'm removing it here. I've tested the BoxSymbol implementation and it does not seem to suffer (still compiles, still runs).


emspin (approved):
Thanks! Had these locally but did not push last week -- sorry about that!

callendorph:
Tests run locally (minus bga regression)

PR #84, opened 2024-07-16
JITX-8858: add support functions to JSL for ethernet connector

emspin:
- add helper in make-1000T
- add VirtualCutouts to VirtualLP

### DIFF
--- a/src/landpatterns/VirtualLP.stanza
+++ b/src/landpatterns/VirtualLP.stanza
@@ -128,6 +128,30 @@ public defmethod make-landpattern (cu:VirtualCopper, pose:Pose = DEF_VLP_LOC) :
   inside pcb-landpattern:
     copper(layer-index(cu)) = pose * shape(cu)
 
+public defstruct VirtualCutout <: VirtualElement :
+  pose:Pose
+  shape:Shape
+
+  ; Virtual Element
+  name?:Maybe<String> with: (as-method => true)
+  class:Vector<String> with: (as-method => true)
+with:
+  constructor => #VirtualCutout
+  printer => true
+
+public defn VirtualCutout (
+  pose:Pose,
+  shape:Shape
+  --
+  name?:Maybe<String> = None(),
+  class:Seqable<String>|String = []
+  ) -> VirtualCutout :
+  #VirtualCutout(pose, shape, name?, to-class-vector(class))
+
+public defmethod make-landpattern (p:VirtualCutout, pose:Pose = DEF_VLP_LOC) :
+  inside pcb-landpattern:
+    layer(Cutout()) = transform(pose * /pose(p), shape(p))
+
 public defstruct VirtualArtwork <: VirtualElement :
   layer-spec:LayerSpecifier
   shape:Shape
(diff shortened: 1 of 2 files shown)


src/landpatterns/VirtualLP.stanza
```diff
@@ -194,6 +218,12 @@ public defstruct VirtualLP <: VirtualElement :
   <DOC>
   metal:Vector<VirtualCopper>
 
+  doc: \<DOC>
+  Virtual `cutout` statements as placced in the virtual landpattern
```
callendorph, line 222: Spelling: `placced` -> `placed`

src/protocols/ethernet/MDI/MDI-1000Base-T.stanza
```diff
@@ -71,6 +71,9 @@ are expected to be passed in the right order - otherwise,
 we may setup the wrong pin assignment constraints.
 <DOC>
 public defn make-1000Base-T-supports (dps:JITXObject ...):
+   make-1000Base-T-supports(dps)
+
+public defn make-1000Base-T-supports (dps:Tuple<JITXObject>):
```
callendorph, line 76: Why do we need this ?  In absence of this function, you should be able to do:

```
make-1000Base-T-supports(dps = [...])
```

In other words - when you use a `variadic` like `dps` above, you can also pass it as a named keyword argument.

emspin: I think it's helpful to have a convenience function to handle the case where the user passes in a tuple without the named keyword argument. Otherwise they'll get 

```
Cannot call function 'make-1000Base-T-supports' with given arguments:
  Argument 0: Passing type 'Tuple<JITXObject>' to expected type 'JITXObject'.
stanza> 
```

callendorph: OK :+1:

src/landpatterns/VirtualLP.stanza
```diff
@@ -128,6 +128,30 @@ public defmethod make-landpattern (cu:VirtualCopper, pose:Pose = DEF_VLP_LOC) :
...
+  shape:Shape
+  --
+  name?:Maybe<String> = None(),
+  class:Seqable<String>|String = []
+  ) -> VirtualCutout :
+  #VirtualCutout(pose, shape, name?, to-class-vector(class))
+
+public defmethod make-landpattern (p:VirtualCutout, pose:Pose = DEF_VLP_LOC) :
+  inside pcb-landpattern:
+    layer(Cutout()) = transform(pose * /pose(p), shape(p))
```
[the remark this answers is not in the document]
emspin: My initial plan was just to have `Shape` sincee most `Shape`s have a `pose` as you mentioned. But not all `Shape`s are guaranteed to have a `pose` so I ended up adding it separately. If there's not really a use case or a viable case where we would have a cutout `Shape` that didn't have a pose then we can get rid of it altogether here. 

Additionally, yes the ordering matters here and is incorrectly applied. But perhaps we can just remove the extra `pose` option altogether if you agree.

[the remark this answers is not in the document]
callendorph: Let's remove the extra pose for now. 

I think @falstjit had been pushing for a consistent `pose` interface for shapes and I tend to agree with that. I would prefer we had consistency in the JITX runtime implementation. 

If that isn't possible or won't be done - then we will need to update this entire interface to support an explicit `pose` - for example for `VirtualCopper`.

src/landpatterns/VirtualLP.stanza
callendorph: This fix is still outstanding

callendorph (approved):
Looks good :+1:

I think the misspelling is still outstanding.

PR #85, opened 2024-07-16
Use lane-pair for PCIe

bhusang:
This PR is just to use of lane-pair in the PCIe code instead of explicit separate RX/TX.


callendorph (approved):
Looks good :+1:
I ran the example PCIe design locally and it routes.

PR #86, opened 2024-07-16
Added offset parameters for Row and Column Major Numbering.

callendorph:
Adds `row-offset` and `col-offset` optional parameters for Row and Column major numbering types. 

I've added tests that show the behavior of each.


emspin (approved):
Great. I like this instead of a single value offset since it kind of forces the user to double check their offset.

PR #87, opened 2024-07-16
Added the cutout interface for the `Package` type.

callendorph:
This adds a default implementation that does nothing because most implementations won't require this.

### DIFF
--- a/src/landpatterns/packages.stanza
+++ b/src/landpatterns/packages.stanza
@@ -319,7 +319,7 @@ public defmulti make-keep-out (
   )
 
 doc: \<DOC>
-Default `make-keep-out` - No Keep-out Regions Applied.
+Default `make-keep-out` - Constructs the Keepouts Only.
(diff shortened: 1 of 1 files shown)


src/landpatterns/packages.stanza
```diff
@@ -331,6 +331,67 @@ public defmethod make-keep-out (
...
+@param pkg Package definition to build
+@param vp Virtual Landpattern Container
+<DOC>
+public defmulti build-cut-out (
+  pkg:Package,
+  vp:VirtualLP,
+  ) -> False
+
+doc: \<DOC>
+Default `build-keep-out` - No cut-out regions created.
```
emspin, line 358: nit: `keep-out` => `cut-out`

emspin (approved):
Small find/replace typo. Otherwise looks good.

callendorph:
Fixed and ran some tests on the examples.

PR #88, opened 2024-07-16
Flesh out QFP

tjknoth:
For use with STM microcontroller library.
This will need some work -- I basically just copy-pasted from QFN. Opening a PR so that somebody more qualified can fix it up.

### DIFF
--- a/src/landpatterns/QFP.stanza
+++ b/src/landpatterns/QFP.stanza
@@ -9,13 +9,7 @@ defpackage jsl/landpatterns/QFP:
   import jsl/ensure
   import jsl/geometry/box
   import jsl/geometry/LineRectangle
-  import jsl/landpatterns/quad
-  import jsl/landpatterns/leads
-  import jsl/landpatterns/packages
-  import jsl/landpatterns/pad-planner
-  import jsl/landpatterns/numbering
-  import jsl/landpatterns/VirtualLP
-  import jsl/landpatterns/silkscreen
+  import jsl/landpatterns/framework
 
 doc: \<DOC>
 QFP Lead Type
@@ -62,16 +56,47 @@ public defn QFP-Lead (
   #QFP-Lead(lead-type, length, width)
 
 doc: \<DOC>
-Helper for Creating QFP Lead Profiles
+Select the QFP protrusion type depending on pitch
+
+@param pitch Distance between adjacent leads on a
+side of a QFP. This value is in mm.
+@return {@link SmallGullWingLeads} or {@link BigGullWingLeads}
+depending on whether pitch is less than 0.625mm or greater,
+respectively.
+<DOC>
+public defn select-QFP-protrusion (pitch:Double) -> LeadProtrusion:
+  if pitch < 0.625:
+    SmallGullWingLeads()
+  else:
+    BigGullWingLeads()
+
+doc: \<DOC>
+Construct the Lead Profile for a Quad Flat Package
+
+This function constructs a `Quad-Lead-Profile` for
+a square package. The protrusion type will be selected
+based on the `pitch` using {@link select-QFP-protrusion}
+
+@param span Lead span for both horizontal and vertical
+edges of the quad package. This value is in mm. This value
+is often labeled as dimension `E` or `D` in a mechanical
+drawing.
+@param pitch Distance between adjacent leads on the side
+of the package. This value is in mm. This value is
+often labeled as dimension `e` in a mechanical drawing.
+@param lead-length Lead foot length of the QFP lead. This
+is often labeled as dimension `L` in mechanical drawings.
+@param lead-width Width of the lead foot. This is often
+labeled as dimension `b` in mechanical drawings.
 <DOC>
 public defn QFP-Lead-Profile (
   --
-  lead-type : LeadProtrusion
   span : Toleranced,
   pitch : Double,
   lead-length : Toleranced,
   lead-width : Toleranced,
   ) -> Quad-Lead-Profile:
+  val lead-type = select-QFP-protrusion(pitch)
   val lead = QFP-Lead(lead-type = lead-type, length = lead-length, width = lead-width)
   val profile = Lead-Profile(span = span, pitch = pitch, lead = lead)
   #Quad-Lead-Profile(profile, None(), profile, None())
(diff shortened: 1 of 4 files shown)


src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
+  import jsl/ensure
   import jsl/design/settings
   import jsl/landpatterns/leads
   import jsl/landpatterns/quad
+  import jsl/landpatterns/packages
+  import jsl/landpatterns/pad-planner
+  import jsl/landpatterns/numbering
 
+doc: \<DOC>
+QFN Lead Type
```
callendorph, line 15: `QFN` -> `QFP`

src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
+<DOC>
+public defstruct QFP-Lead <: SMT-Lead:
+  lead-type:LeadProtrusion with: (
+      as-method => true,
+      default => SmallGullWingLeads(), ; TODO: is this OK?
```
callendorph, line 20: Fine for now - see the `SOIC.stanza` file for how I callout specifics on what lead profile to use.

src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
+    )
+  width:Toleranced with: (
+    as-method => true,
+    ensure => ensure-positive!
+    )
+with:
+  equalable => true
+  hashable => true
+  printer => true
+  constructor => #QFN-Lead
```
callendorph, line 34: `QFN-Lead` -> `QFP-Lead`

src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
-    num-pins, pitch,
-    lead-length, lead-width,
-    package-size, ; lead-span
-    lead-type,
-    package-size,
-    density-level = density-level,
-    corner-pads? = corner-pads?
-    thermal-pad? = thermal-pad?
-  )
+  #QFN-Lead(lead-type, length, width)
```
callendorph, line 45: `QFN`  -> `QFP`

src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
+public defstruct QFP <: Quad-Package :
+  doc: \<DOC>
+  Number of leads ignoring any depopulated leads and thermal leads
+  <DOC>
+  num-leads:Int with: (ensure => ensure-even-positive!, as-method => true)
+
+  doc: \<DOC>
+  Lead Profiles for all of the QFP's edges
+
+  By default the QFN uses the {@link SmallGullWingLeads} protrusion
```
callendorph, line 71: `QFN` -> `QFP`

I think you want this to talk about using the `QFP-Lead` type as the base profile.

src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
+  defn to-deci (v:Double) -> String:
+    val v* = to-int( v * 100.0 )
+    to-string("%_" % [v*])
+  val body = package-body(pkg)
+  val p = to-deci(pitch $ x-leads $ lead-profile(pkg))
+  val w = to-deci(typ $ width(body))
+  val l = to-deci(typ $ length(body))
+  val h = to-deci(typ $ height(body))
+  val n = num-leads(pkg) + (1 when thermal-lead?(pkg) is-not False else 0)
+  to-string("QFP%_P%_X%_X%_-%_N" % [p, w, l, h, n])
```
callendorph, line 138: IPC7351:

Screenshot 2024-07-16 at 4 00 52 PM

I think the naming deviates - I can help get more explicit examples

src/landpatterns/QFP.stanza
```diff
@@ -3,55 +3,190 @@ defpackage jsl/landpatterns/QFP:
...
+
+public defmethod courtyard-excess (pkg:QFP) -> Double :
+  val protrusion = lead-type $ lead $ x-leads $ lead-profile $ pkg
+  val fillets = lead-fillets(protrusion, density-level = density-level(pkg))
+  courtyard-excess(fillets)
+
```
callendorph, line 144: We will need to add the `build-silkscreen` similar to what the QFN is doing.

We will want the outline to be inside the pads unlike in QFN where it is outside the pads.

src/landpatterns/QFP.stanza
tjknoth: what does "lead span l1" and "lead span l2 nominal" mean here?

callendorph: The lead span is from the edge of the pin on one side to the edge of the pin on the other side. 

In the following image - there are unequal numbers of pins on either side:
image

Leadspan 1 is the longer leadspan (typically) and lead span 2 is the shorter one.

tjknoth: The naming convention in the screenshot above differs from the datasheet, where they just use `LQFP100` in this case (100-pin, low-profile). Is there a general pattern or is this manufacturer-specific or something?

tjknoth:
Marking this as ready for review. I believe all feedback has been addressed. Naming convention seems in line with the Altium examples I've seen (though it may still be incomplete -- I only have prefix support for "thin" and "small" packages). Silkscreen should respect pad clearance, but it's done so pretty conservatively.

src/landpatterns/QFP.stanza
callendorph: The naming is often manufacturer specific and depends on process and features. Common names are:

1.  `LQFP` - low profile
2.  `TQFP` - thin
3.  `CQFP` - Ceramic
4.  `PQFP` - Plastic

Some of this is also a branding thing. Let's not worry to much about it. The user can override the `name` method if they want to change it to something else.

tjknoth: Sounds good -- I just implemented "T" and "S" since those were in both sources you gave me.

callendorph:
OK - I merged in main and then made some updates to the QFP code. I then added an example for checking the generator. 

Screenshot 2024-07-30 at 9 36 00 PM

Notice that the body outline is over the top of the pads. I think we should update the generator so that it 
finds the inner most edge of each of the pads and then offsets from that edge to create a rectangle. Then we should compare that rectangle with the package body rectangle and take the smaller of the two rectangles to create the silkscreen outline. 

I used the datasheet from [here](https://www.digikey.com/htmldatasheets/production/1334005/0/0/1/ep1c3t100c6.html) - Page 53, 44pin TQFP:

Screenshot 2024-07-30 at 9 46 54 PM

Screenshot 2024-07-30 at 9 47 04 PM

tjknoth:
OK -- just pushed some updates to the silkscreen computation. It now uses your suggested algorithm (minimum of package outline and pad interior adjusted for clearance).
Might rewrite the computation if I can find a cleaner way, but for now it works.
Results:
Screenshot 2024-08-01 at 10 45 40 AM
Screenshot 2024-08-01 at 10 45 13 AM
Generated the first screenshot by bumping up clearance just to show that it gets respected. Second screenshot uses default clearance parameter.

callendorph:
>Generated the first screenshot by bumping up clearance just to show that it gets respected. Second screenshot uses default clearance parameter.

Are you generating the clearance with respect to the copper or to the soldermask?

In the second image, it looks like the silkscreen is on top of the soldermask without any clearance.

tjknoth:
Fixed -- was using the wrong `bounds` function.

callendorph (approved):
Looks good - Nice work! :+1:

PR #89, opened 2024-07-16
full i2s spec

tjknoth:
Extends bundle with full i2s spec -- need 5 pins to allow master mode. Motivated by STM microcontroller, which supports both modes.
`i2s-minimal()` is the simple interface, `i2s-master` requires 5 pins.
Interfaces probably need reworking, docs need EE knowledge. Added a reference to something that seems to specify the richer protocol.

### DIFF
--- a/src/bundles/comms.stanza
+++ b/src/bundles/comms.stanza
@@ -311,7 +311,51 @@ specification is present in NXP's references
 @member ws Word Select - typically for left vs right differentiation
 @member sd Data Signal
 <DOC>
-public pcb-bundle i2s :
+
+public pcb-enum jsl/bundles/comms/I2SPins :
+  I2S-MCK 
+  I2S-SDI
+
+pcb-bundle i2s-b (pins:Collection<I2SPins>):
   pin sck
   pin ws
-  pin sd
+  pin sdo
+  for p in pins do :
+    switch(p) :
+      I2S-MCK : mak-pin(`mck)
+      I2S-SDI : mak-pin(`sdi)
+
+doc: \<DOC>
+Inter-Integrated Sound - I2S Protocol
+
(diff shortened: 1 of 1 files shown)


src/bundles/comms.stanza
```diff
@@ -311,7 +311,51 @@ specification is present in NXP's references
 @member ws Word Select - typically for left vs right differentiation
 @member sd Data Signal
 <DOC>
-public pcb-bundle i2s :
+
+public pcb-enum jsl/bundles/comms/I2SPins :
+  I2S-MCK
```
callendorph, line 316: `MCK` is not part of the protocol specification. There is no reason it must come from the microcontroller. It could just as easily come from a standalone oscillator.

I don't think this should be included.

src/bundles/comms.stanza
```diff
@@ -311,7 +311,51 @@ specification is present in NXP's references
...
+public pcb-enum jsl/bundles/comms/I2SPins :
+  I2S-MCK 
+  I2S-SDI
+
+pcb-bundle i2s-b (pins:Collection<I2SPins>):
   pin sck
   pin ws
-  pin sd
\ No newline at end of file
+  pin sdo
```
callendorph, line 322: The `SDO` vs `SDI` distinction here is because of the microcontroller. Not because of the protocol. 

Unfortunately - the bundle language in JITX here is insufficient to actually describe the system we need. 

I don't think we should do this. 

I propose that we:
1.  We should instead leave it as a 3-pin bundle per the specification.  
2. We don't add the I2S supports bundle to the STM32 or whatever MCU yet. 

This way the user can create a `I2S` supports statement that meets their needs instead of us trying to force a solution that will never work properly because the language is deficient.

PR #90, opened 2024-07-16
Jsl/add cutouts to virtual elements pass

emspin:
Required for cutouts to actually get generated properly :)


callendorph (approved):
:+1: Thanks for adding the test!

PR #91, opened 2024-07-17
PROD-313: Fix property checking in BoxSymbolv2

emspin:
Property check was broken as the value of wildcard type was not properly accessed for comparison.


tjknoth (approved):
Just tested -- works great!

PR #93, opened 2024-07-19
revert the last commit after Carl's approval of v1

lchao-jitx:
Revert the last commit after Carl's approval of v1.

That last commit has the following changes:
  - enhanced `make-layer-stack` to support multiple adjacent dieletric layers like `[copper [prepreg1 prepreg2]]`, which is found in many JLC-PCB stackup.
  - Specific references to JLC-PCB stackups like jlcpcb-2313 was removed so the examples are general

These changes are reverted.


callendorph (approved):
Thank you!

emspin (approved):
Thanks Liang! Confirmed that `jstanza run-test jsl/tests/layerstack` passes successfully.

PR #94, opened 2024-07-19
Syntax Reform Updates for `layer()`

callendorph:
This PR needs to merge in sequence with the JITX 3.14 (PI) release which will have these syntax reforms.


bhusang (approved):

callendorph:
We have to split this up and do a different way because of symbol name issues. New PR as soon as #98 merges.

PR #95, opened 2024-07-22
JSL: Revert VirtualCutout, augment artwork

emspin:
This diff reverts unnecessary addition of VirtualCutout and adds helper functions to create `Cutout()` and `ForbidCopper()` regions using `VirtualArtwork`

### DIFF
--- a/src/landpatterns/VirtualLP.stanza
+++ b/src/landpatterns/VirtualLP.stanza
@@ -128,28 +128,6 @@ public defmethod make-landpattern (cu:VirtualCopper, pose:Pose = DEF_VLP_LOC) :
   inside pcb-landpattern:
     copper(layer-index(cu)) = pose * shape(cu)
 
-public defstruct VirtualCutout <: VirtualElement :
-  shape:Shape
-
-  ; Virtual Element
-  name?:Maybe<String> with: (as-method => true)
-  class:Vector<String> with: (as-method => true)
-with:
-  constructor => #VirtualCutout
-  printer => true
-
-public defn VirtualCutout (
-  shape:Shape
-  --
-  name?:Maybe<String> = None(),
-  class:Seqable<String>|String = []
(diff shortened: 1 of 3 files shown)


src/landpatterns/VirtualLP.stanza
```diff
@@ -393,6 +363,33 @@ public defn add-artwork (
...
+  --
+  name:String = ?,
+  class:Seqable<String>|String = []
+  ) -> False:
+  append(vp, VirtualArtwork(Cutout(), shape, name? = name, class = class))
+
+doc: \<DOC>
+  Add keepout regions to the landpattern with a sequence of given shapes.
+<DOC>
+public defn add-keepouts (
```
callendorph, line 381: Why make `add-cutout` singular and `add-keepouts` plural with a sequence ? 

These inconsistencies are not good as part of a public facing API. 

Check the `add-copper` implementation - it has polymorphics that support both. This is the way these interfaces should be added.

src/landpatterns/packages.stanza
```diff
@@ -363,35 +363,6 @@ public defmethod build-cut-out (
...
-public defmethod make-cut-out (
-  pkg:Package
-  --
-  pose:Pose = DEF_LP_POSE
-  ):
-  inside pcb-landpattern:
-    val virt = VirtualLP(pose)
-    build-cut-out(pkg, virt)
-    make-landpattern(virt)
-
```
callendorph, line 394: Let's keep this part. I think adding this to the `Packages` interface is fine.

callendorph (approved):
1.  Let's make the interfaces consistent like `add-copper`.
2.  Let's keep the changes to `Package` - I think those interfaces are still work keeping. 

Other than that, looks good :+1:

callendorph:
Please run the unit tests before merging.

PR #97, opened 2024-07-23
Added methods to support pin extraction for 2-pin devices

callendorph:
This adds a function for checking if a 2-pin component is the `p[1]/p[2]` version or the `c/a` version and returning the pins in a known order.

### DIFF
--- a/src/design/introspection.stanza
+++ b/src/design/introspection.stanza
@@ -47,11 +47,11 @@ public defn get-port-by-name (obj:JITXObject, name:String) -> Maybe<JITXObject> 
     return(None())
 
 doc: \<DOC>
-Get the Cathode / Anode Pins of a device if they exist
+Get the Cathode / Anode Ports of a device if they exist
 @param obj Two-pin component or module instance that may or
-may not be a polarized device with cathode (`c`) and anode (`a`) pins.
+may not be a polarized device with cathode (`c`) and anode (`a`) ports.
 @return If the passed object is a polarized device, then this returns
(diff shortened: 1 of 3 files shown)


src/design/introspection.stanza
```diff
@@ -23,4 +23,50 @@ public defn get-instantiable (comps:JITXObject|InstantiableType) -> Instantiable
...
+
+Depending on whether the passed device is polarized.
+
+<DOC>
+public defn get-pins (obj:JITXObject) -> [JITXObject, JITXObject] :
+  val mCA = get-cathode-anode(obj)
+  match(mCA):
+    (oCA:One<[JITXObject, JITXObject]>):
+      val ret = value(oCA)
+      ret
```
bhusang, line 69: Why not just:
```
  match(mCA):
    (oCA:One<[JITXObject, JITXObject]>:
      value(oCA)
```

bhusang (approved):

src/design/introspection.stanza
callendorph: I think this was a debugging thing at one point. I can simplifyt

callendorph:
I've renamed this function to `get-element-ports` so that we don't reference `pins` and we make it more clear that it is about lumped circuit elements like resistors, capacitors, etc. I wanted to avoid something like `get-2-pin-ports` because that seemed even more confusing. 

If you have a better name for any generic 2-pin lumped circuit element device - that would be great. I don't like referring to them as `TwoPin` as the symbol code does.

PR #99, opened 2024-07-24
Syntax Reforms PR (merge after 3.14 is released)

callendorph:
This PR includes work from both me and @bhusang to update the code with the reformed syntax that should land in `3.14` ( 🤞 )

This should be the common place where to handle all syntax reforms. If subsequent PRs create or edit those lines - we should modify them here so that we can fix everything with one single PR. 

This PR must NOT merge until 3.14 is released.


bhusang (approved):
LGTM

PR #100, opened 2024-07-25
Virtual LP FIxups to Support ethernet connector

callendorph:
1.  Support `build-all` overrides better
2.  Tools for easier child node lookup by name
3.  Bugfix for `get-pads` that fixes the courtyard outline for LPs with multiple child nodes
4.  Added silkscreen from package body helper routine.

### DIFF
--- a/src/landpatterns/packages.stanza
+++ b/src/landpatterns/packages.stanza
@@ -470,7 +470,13 @@ public defmulti build-all (
   virt:VirtualLP
   ) -> False
 
-public defn package-build-all-seq (
+doc: \<DOC>
+Externally Defined Build All Order
+
+This function can be used when you want to override the
+default behavior of default `build-all` defmethod.
+<DOC>
+public defn package-build-all-order (
   pkg:Package,
   virt:VirtualLP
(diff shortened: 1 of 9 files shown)


src/landpatterns/packages.stanza
```diff
@@ -470,6 +470,17 @@ public defmulti build-all (
   virt:VirtualLP
   ) -> False
 
+public defn package-build-all-seq (
```
emspin, line 473: qq: what's the `seq` suffix here referring to?

callendorph: Sequence - ie, these functions all get called in a particular order. 

I could replace it with `order` if you think that would be less confusing (ie, not confused with list operator `seq`)

emspin: Yeah, that's probably preferred. I had assumed it referred to the order in which things were being built, as opposed to the operator, but explicitly dodging the possible ambiguity is better.

emspin (approved):
Only comment is to rename function to not clash with term `seq`

Thanks for making the changes!

PR #105, opened 2024-07-25
PROD-250: Better error reporting for symbols with banked and unbanked

emspin:
Improve exception messages when there are pins with and without `bank` property

### DIFF
--- a/src/symbols/box-symbol.stanza
+++ b/src/symbols/box-symbol.stanza
@@ -342,26 +342,21 @@ defn save-pin-info (x:BoxSymbol, bank:Int|Ref|False = false) -> BoxSymbolBank :
   val decorators = HashTable<Ref,DecoratorInfo>()
 
   val sorted-pins = sort-by-row-index(pins(obj(x)))
-  val banked-pins = Vector<JITXObject>()
-  val unbanked-pins = Vector<JITXObject>()
 
-  for p in sorted-pins do :
-    match(get-property?(p, BANK_PROP)) :
-      (o:One):
-        add(banked-pins, p)
-      (n:None):
-        add(unbanked-pins, p)
+  val [banked, unbanked] = for p in sorted-pins split :
+    get-property?(p, BANK_PROP) is One
 
-  if length(banked-pins) > 0 and length(unbanked-pins) > 0 :
+  if not empty?(banked) and not empty?(unbanked) :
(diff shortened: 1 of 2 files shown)


src/symbols/box-symbol.stanza
```diff
@@ -342,23 +342,29 @@ defn save-pin-info (x:BoxSymbol, bank:Int|Ref|False = false) -> BoxSymbolBank :
...
-        false
-      (f:False, o:One) :
-        false
-      (f1:False, f2:None) :
-        add(filtered-pins, p)
+    match(get-property?(p, BANK_PROP)) :
+      (o:One):
+        add(banked-pins, p)
+      (n:None):
+        add(unbanked-pins, p)
```
callendorph, line 353: I think you can do this with `split`:

```
val [banked, unbanked] = for p in sorted-pins split:
   has-property?(dot(p, BANK_PROP))
```

callendorph:
I first rebased this branch on main to fix the BGA pad regression. Unit tests pass so that is good. 

I made this change:

```
diff --git a/examples/symbols/box-symbol.stanza b/examples/symbols/box-symbol.stanza
index 5c3fc46..74f9745 100644
--- a/examples/symbols/box-symbol.stanza
+++ b/examples/symbols/box-symbol.stanza
@@ -440,7 +440,7 @@ pcb-component test-up-column:
     [rail.V+ | p[6] | Right | 0 | "VCC" | test ]
     [rail.V- | p[2] | Left | 1 | "VSS" | test ]
     [ADDR[0] | p[5] | Right | 1 | "A0" | test ]
-    [ADDR[1] | p[4] | Right | 1 | "A1" | test  ]
+    [ADDR[1] | p[4] | Right | 1 | "A1" | -  ]
 
   val box = BoxSymbol(self, params = BoxSymbolParams(
     show-grid = true
(END)
```

Then I ran the design and I see this error message:

Screenshot 2024-07-26 at 8 38 43 AM

I'm still not seeing the name of the pins that don't have bank assignments. Can we please make it so that we see the pin names and not the object references ?

emspin:
Forgot to apply `ref`. Updated tests and also ran your example. Should be good now.

callendorph (approved):
:+1: Looks good.
I ran locally and saw good names - Nice work!

PR #106, opened 2024-07-26
[Bugfix] Fixed the antipad not allowing `Bottom` as the side.

callendorph:
I had mistakenly thought that the `Bottom` wasn't usable because I didn't know the size of the board from the code. But `ForbidCopper` has an interface that allows it. I've removed the checks and I'm just using that directly.


falstjit (approved):
Looks good to me :+1:

PR #107, opened 2024-07-26
Convert Courtyard LineRectangle to Rectangle to make more sane

callendorph:
This addresses the JSL short coming of this:

https://linear.app/jitx/issue/PROD-377/courtyard-outline-undocumented-behavior

### DIFF
--- a/src/landpatterns/courtyard.stanza
+++ b/src/landpatterns/courtyard.stanza
@@ -62,8 +62,7 @@ public defn compute-courtyard-boundary (
   val pkg-outline = bounds(envelope(pkg-body, density-level = density-level))
   val outline = union(pad-outline, pkg-outline)
   val outline* = fatten(excess, outline)
-  pos * LineRectangle(outline*, line-width = DEF_COURTYARD_WIDTH)
-
+  pos * to-Rectangle(outline*)
 
 doc: \<DOC>
 Generator for Courtyard Boundary
--- a/src/landpatterns/courtyard.stanza
+++ b/src/landpatterns/courtyard.stanza
@@ -12,18 +12,18 @@ defpackage jsl/landpatterns/courtyard:
   import jsl/landpatterns/VirtualLP
   import jsl/landpatterns/packages with:
     only => (PackageBody, envelope)
-  import jsl/geometry/LineRectangle
 
-public val DEF_COURTYARD_WIDTH = 0.1
+public val DEF_COURTYARD_LINE_WIDTH = 0.1
 
 public defn build-courtyard-origin (
   vp:VirtualLP,
   --
+  line-width:Double = DEF_COURTYARD_LINE_WIDTH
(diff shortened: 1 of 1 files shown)


src/landpatterns/courtyard.stanza
```diff
@@ -62,8 +62,7 @@ public defn compute-courtyard-boundary (
   val pkg-outline = bounds(envelope(pkg-body, density-level = density-level))
   val outline = union(pad-outline, pkg-outline)
   val outline* = fatten(excess, outline)
-  pos * LineRectangle(outline*, line-width = DEF_COURTYARD_WIDTH)
```
falstjit, line 65: perhaps getting rid of the DEF_COURTYARD_WIDTH constant would be a good idea at some point? Or perhaps it's needed for things that aren't the outline? The width of the outline can technically be controlled through UI styles.

falstjit (approved):

src/landpatterns/courtyard.stanza
callendorph: I can at least make it more specific to the origin marker.

falstjit: is it common practice to have the origin marker on the courtyard layer? perhaps we need to be smarter in the UI for stuff like that (e.g. for alignment)?

callendorph: It is pretty common in my experience. The default IPC footprint generator in Altium will create an origin marker like this. I tend to like having them for components where the land pattern center is not the same as the "center of mass" of the component.

falstjit: Ideally we'd create our landpatterns so their origin is the actual "origin" that we want, same for modules, but I don't know if that's always feasible especially for imports.

callendorph: Unfortunately it isn't always that easy. For example - see this part:
https://productfinder.pulseelectronics.com/api/open/part-attachments/datasheet/JD0-0001NL

Screenshot 2024-07-26 at 12 10 05 PM

Notice how the mechanical drawing references all of the dimensions from the mechanical mounting hole as a datum ? 

In legacy CAD - what you will often do is either create the landpattern in that shift frame of reference and then re-set the origin once you are done or  you will just leave it in the shifted frame of reference. 

In the landpattern framework in JSL - I have an implementation that allows me to create a child node in that shift frame of reference. The JSL approach is far superior in my opinion because I can work in the shifted frame of reference but keep the origin of the landpattern at the components center of mass.

falstjit: Yeah, makes sense. We don't really care for the details of the land pattern construction at all, it's the components frame of reference that's relevant for things like placement and what not. I think what I'm saying is what you're also saying, that it would be cool if the "origin marker" isn't just something in a courtyard layer, but the actual origin of either module or component, I feel like that'll make things easier in the long run.

PR #108, opened 2024-07-26
Add a new Grid management facility for landpattern generation

callendorph:
The goal is to:

1.  Reduce duplication in the grid generation. 
2.  Add routines that can support the ethernet connector generator.

This also includes several bugfixes. 

Unit tests pass when I run them locally.

# Notes

1.  This changes the `BGA` interface to put the `rows/columns` management into the BGA Planner instead of on the BGA itself.
    1.  This seems to jive well with the Corner Cut planner from @alijitx 
2.  I've updated dual-row and quad to use GridPlanner for their construction mechanism.
    1.  This works well for both.
    2.  But the helper function only works for the dual not the quad, because of the two separate lead-profile mechanisms the quad supports.
3.  I've updated the headers functionality to use this pad grid concept and added examples of using the headers.

# Future Work 

1.  We need better handling for through-holes. 
     1.  Current implementation creates IPC-based hole and pad sizes. 
     2.  We should give the user the ability to override both 
2.  We need to support implement some ability to do pin-1 pad planners for through-holes.
3.  Need to test generalized stagger grid support for ethernet connector.
4.  We need some rework in the through hole (axial/radial) components to use pad-grid.

### DIFF
--- a/src/landpatterns/BGA/package.stanza
+++ b/src/landpatterns/BGA/package.stanza
@@ -173,22 +173,32 @@ public defn build-vpads (
   val pad-size = Dims(lead-diam(pkg), lead-diam(pkg))
   ; Generate the grid
   val planner = get-bga-planner(pkg)
-  defn gen-pad-info () -> Seq<VirtualPad> :
-    for pos in grid(planner) seq?:
-      val [r, c] = [row(pos), column(pos)]
-      val pad-id = to-pad-id(lead-numbering(pkg), r, c)
-      val pad-gen? = pad-generator(planner, r, c)
-      match(pad-gen?):
-        (_:False): None()
-        (pad-gen):
-          val cls = [
-            "pad",
-            to-string("col-%_" % [c]),
-            to-string("row-%_" % [r])
-          ]
-          One $ VirtualPad(pad-id, pad-gen(pad-size), pose(pos), class = cls)
+  ; defn gen-pad-info () -> Seq<VirtualPad> :
+  ;   for pos in grid(planner) seq?:
+  ;     val [r, c] = [row(pos), column(pos)]
+  ;     val pad-id = to-pad-id(lead-numbering(pkg), r, c)
+  ;     val pad-gen? = pad-generator(planner, r, c)
+  ;     match(pad-gen?):
+  ;       (_:False): None()
+  ;       (pad-gen):
+  ;         val cls = [
+  ;           "pad",
+  ;           to-string("col-%_" % [c]),
+  ;           to-string("row-%_" % [r])
+  ;         ]
+  ;         One $ VirtualPad(pad-id, pad-gen(pad-size), pose(pos), class = cls)
 
-  append-all(vp, gen-pad-info())
+  ; append-all(vp, gen-pad-info())
+
+  val pad-seq = pad-grid-smt(
+    pad-size = pad-size,
+    planner = planner,
+    lead-grid = planner,
+    num-scheme = lead-numbering(pkg)
+  )
+
+  append-all(vp, pad-seq)
+
 
 public defmethod build-pads (
   pkg:BGA,
--- a/src/landpatterns/BGA/package.stanza
+++ b/src/landpatterns/BGA/package.stanza
@@ -15,6 +15,7 @@ defpackage jsl/landpatterns/BGA/package:
   import jsl/landpatterns/framework
   import jsl/landpatterns/BGA/pads
   import jsl/landpatterns/BGA/planner
+  import jsl/landpatterns/grid-planner
 
 doc: \<DOC>
 Ball Grid Array Package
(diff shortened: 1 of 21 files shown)


callendorph:
You may want to hold off reviewing this - I need to add some more things to make it usable.

src/landpatterns/pad-planner.stanza
```diff
@@ -168,30 +198,50 @@ public defmethod th-pad-generator (x:PadPlanner, row:Int, column:Int) -> ( (Dims
...
 
-TODO - Should we make this return `Circle|Capsule` instead of `Shape` ?
-  Holes in the board generally speaking are very expensive to make in
-  other shapes but I don't necessarily want to limit things here.
-  Perhaps with make non-Circle/Capsule shaped holes a check that is
-  disable-able so that the user can override.
 <DOC>
+
+; TODO - Should we make this return `Circle|Capsule` instead of `Shape` ?
+;   Holes in the board generally speaking are very expensive to make in
```
emspin, line 240: If the enterprise customers can afford it then I would  suggest keep it `Shape`

src/landpatterns/BGA/package.stanza
```diff
@@ -115,32 +92,58 @@ public defn BGA (
...
-  pad-planner:BGA-PadPlanner = Full-Matrix-Planner(),
+  pad-planner:BGA-PadPlanner = Full-Matrix-Planner(
+    rows = rows,
+    columns = columns,
+    pitch = pitch
+    ),
   lead-numbering:Numbering = Grid-Numbering(rows, columns)
   density-level:DensityLevel = DENSITY-LEVEL
 ) -> BGA :
   #BGA(
```
emspin, line 103: For this first constructor, I wonder if we should just omit `pad-planner` and `lead-numbering` as arguments and just hardcode the defaults. `rows` and `columns` only have any use with the defaults and to me it would help make the difference between the constructors more apparent.

```
public defn BGA (
  --
  num-leads:Int,
  rows:Int,
  columns:Int,
  lead-diam:Double,
  pitch:Double|Dims,
  package-body:PackageBody,
  density-level:DensityLevel = DENSITY-LEVEL
) 
...
public defn BGA (
  --
  num-leads:Int,
  lead-diam:Double,
  package-body:PackageBody,
  pad-planner:BGA-PadPlanner,
  lead-numbering:Numbering = Grid-Numbering(rows(pad-planner), columns(pad-planner))
  density-level:DensityLevel = DENSITY-LEVEL
) -> BGA
```

callendorph: That's fair. I think I was primarily trying to preserve the existing interface so as not to break any existing code. 
But big picture - we're breaking a lot of interfaces already so its not a big deal.

src/landpatterns/pad-planner.stanza
callendorph: We can keep the interface as `Shape`.

src/landpatterns/dual-row.stanza
```diff
@@ -291,3 +273,95 @@ public defn build-smd-pin-1 (
   add-artwork(vp, Silkscreen("pin-1-marker", side), Circle(marker-pos, line-width), name = "pin-1-dot", class = "pin1-marker")
 
 
+doc: \<DOC>
+Dual-Row Grid Planner
```
emspin, line 277: Probably lack of EE context on my side, but wouldn't `Dual-Column` be more intuitive here? This also explains my observations when implementing the terminal blocks...

src/landpatterns/numbering.stanza
```diff
@@ -183,6 +183,16 @@ numbering scheme. For example - a `col-offset = 1` would construct:
     8   12
   ```
 
+@param col-skip Value to divide the raw column value by. This parameter
+is useful when working with staggered grids where a pad is active every
+other column. When this value is greater than 1 - then the numbering scheme
+will accept twice as many columns as normal.
```
emspin, line 189: Perhaps a small example would help solidify the explanation here?

I'm assuming "will accept twice as many as normal" means that the numbering scheme would normally not accept values higher than `num-cols` or `num-rows` but if a skip parameter is passed in, then it'll adjust for that and know to allow up to `<skip-value> * num-cols/rows`

callendorph: Ya - I agree that a diagram and an example would be helpful.

> I'm assuming "will accept twice as many as normal" means that the numbering scheme would normally not accept values higher than num-cols or num-rows but if a skip parameter is passed in, then it'll adjust for that and know to allow up to <skip-value> * num-cols/rows

Yes

src/landpatterns/dual-row.stanza
callendorph: `Dual Row` connectors is a pretty common term of art. 

Try google [dual row connector](https://www.google.com/search?q=dual+row+connector
) vs [dual column connector](https://www.google.com/search?q=dual+column+connector). There is a night and day difference.

src/landpatterns/numbering.stanza
```diff
@@ -183,6 +183,16 @@ numbering scheme. For example - a `col-offset = 1` would construct:
...
+
+@param row-skip Value to divide the raw row value by. This parameter
+is useful when working with a staggered grid where a pad is active every
+other row. When this value is greater than 1 - then the numbering scheme
+will accept twice as many rows as normal.
+
 @throws ValueError If num-pads is not divisible by num-columns.
 <DOC>
 public defstruct Column-Major-Numbering <: Numbering :
```
emspin, line 198: lol I can't comment on unchanged lines -- what are your thoughts on just requiring a single offset value? I guess you probably have a much better idea of the common use case but when I used this I ended up having to calculate the row/col offsets before passing them in here, when I already just had a hard value offset that I wanted the pad-ids to start at. i.e. I wanted to start the counting at 20 and just provide that as a starting id. 

Just my 2 cents. If you think users are going to use this more from the perspective of offsetting by rows and columns then ignore my babbling

emspin:
I probably didn't get to take a close look at everything but overall good refactoring and improvements here. Left various thoughts/comments, but nothing blocking for merge.

emspin (approved):

src/landpatterns/numbering.stanza
callendorph: > I guess you probably have a much better idea of the common use case but when I used this I ended up having to calculate the row/col offsets before passing them in here, when I already just had a hard value offset that I wanted the pad-ids to start at. i.e. I wanted to start the counting at 20 and just provide that as a starting id.

I guess I don't understand. I needed to do the exact same thing in the ethernet connector and I could just pass `col-offset = 10` or whatever and it did exactly what i needed to do. There is nothing forcing you to pass in a `row-offset` if you don't need to. 

```
    num-scheme = Row-Major-Numbering(2, 1, col-offset = 12)
```

I don't see any reason to remove.

callendorph:
Unit tests are passing locally. I think build is still not working in the cloud for whatever reason.

PR #117, opened 2024-08-05
Add a thermal via generator (merge after 3.15 is released)

callendorph:
* Adds some instropection utilities. 
* Adds a new thermal via grid generator tool
   *  This is currently a static grid and doesn't have any mechanism to just pack the thermal pad with vias.
   *  We need to implement [changes to the API to add port introspection](https://linear.app/jitx/issue/PROD-419/add-port-introspection-functions) to be able to do this.

This PR will likely be usable in the 3.15 release of JITX. We may want to hold off merging this until that release lands.

### DIFF
--- a/src/design/introspection.stanza
+++ b/src/design/introspection.stanza
@@ -4,6 +4,7 @@ defpackage jsl/design/introspection:
   import jitx
   import jitx/commands
 
+  import jsl/errors
 
 doc: \<DOC>
 Retrieve the `Instantiable` that generates the instances of the passed object.
@@ -89,3 +90,36 @@ public defn get-element-ports (obj:JITXObject) -> [JITXObject, JITXObject] :
     (_:None):
       [obj.p[1], obj.p[2]]
 
+doc: \<DOC>
+Get the net associated with a module or component port
+
+@param pt Port of a component or module.
+@return If a connection is found - this function returns the Net as
+a JITXObject. If no connection is found - then None().
+<DOC>
+public defn get-connected-net (pt:JITXObject) -> Maybe<JITXObject>:
+  inside pcb-module:
+    for n in nets(self) first :
+      if connected?([pt, n]):
+        One(n)
+      else:
+        None()
+
+doc: \<DOC>
+Get the net associated with a module or component port
+@param pt Port of a component or module.
+@return A `Net` object as a JITXObject if the port is connected.
+@throws ValueError if no net connections are found or if more than one net
+connection is found.
+<DOC>
+public defn get-connected-net! (pt:JITXObject) -> JITXObject:
+  inside pcb-module:
+    val n-set = to-tuple $ for n in nets(self) filter :
+      connected?([pt, n])
+
+    if length(n-set) == 0:
+      throw $ ValueError("Invalid Port - No net connection found")
+    else if length(n-set) > 1 :
+      throw $ ValueError("Invalid Port - Multiple Nets Associated - Expected Singular Net")
+
+    n-set[0]
--- a/src/landpatterns/thermal-vias.stanza
+++ b/src/landpatterns/thermal-vias.stanza
@@ -1,0 +1,148 @@
+#use-added-syntax(jitx)
+defpackage jsl/landpatterns/thermal-vias:
+  import core
+  import jitx
+  import jitx/commands
+
+  import jsl/ensure
+  import jsl/errors
+  import jsl/design/introspection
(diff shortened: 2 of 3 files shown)


src/design/introspection.stanza
```diff
@@ -89,3 +90,36 @@ public defn get-element-ports (obj:JITXObject) -> [JITXObject, JITXObject] :
...
+@return If a connection is found - this function returns the Net as
+a JITXObject. If no connection is found - then None().
+<DOC>
+public defn get-connected-net (pt:JITXObject) -> Maybe<JITXObject>:
+  inside pcb-module:
+    for n in nets(self) first :
+      if connected?([pt, n]):
+        One(n)
+      else:
+        None()
```
bhusang, line 106: Philosophically, shouldn't these introspection functions (in this file) be part of jitx-client? I have no issue with them in JSL, but it seems to me that we are going down the path like ocdb where very useful functions ended up in user code versus the client.

src/landpatterns/thermal-vias.stanza
```diff
@@ -0,0 +1,148 @@
...
+  import jsl/design/introspection
+  import jsl/landpatterns/grid-planner
+
+doc: \<DOC>
+Thermal Via Interface
+
+This interface outlines the functions needed to create a thermal via
+grid. The idea is that the `make-thermal-vias` function gets called
+as a generator inside a `pcb-module` definition and targets one of
+the pads of a component instantiated in that module.
```
bhusang, line 18: Should we confirm that the component in question is also placed in code?

bhusang (approved):
Some questions:
1) which introspection functions belong in JSL?
2) should the function confirm that the component in question is code-placed?

callendorph:
> should the function confirm that the component in question is code-placed?

Is there a way to do that ? We would probably have to add to the API. 

One problem I had was that I had a structure like this:

```
pcb-component component:
   mpn = "KSZ9563"

pcb-module wrapper:
  inst C : component
  ; Forgot to add place() statement here 

pcb-module circuit: 
   inst A  : wrapper
   place(A) at loc(0, 0) on Top

```

And that really mucked things up. It is almost like I would need to recursively go from top module down to the component and determine if it is statically placed.

src/landpatterns/thermal-vias.stanza
callendorph: Thats a good idea.

src/design/introspection.stanza
callendorph: Ya - thats fair. 

I think prior to the query API - the port/pad definitions for `p[1]/p[2]` and `c/a` were in OCDB and so it didn't make sense to put them in JITX proper. 

I think with the refactor @plasmajitx  has done - it may now make more sense to put some of those types of functions in JITX. 

There are others here that probably should be in JITX. 

I think we need to talk about organization of that content - I can create a ticket.

bhusang:
I think that we can retrieve the pose of an instance. If we do that, maybe it's enough to confirm that it's at (0.0, 0.0) before instantiating the thermal vias? Or there might be a better way to do that? Ask @tjknoth?

callendorph:
Attempting: 

```
  public inst netsw : ethernet_io_lib/components/KSZ9563/module
  place(netsw) at loc(0.0, 0.0) on Top

  println("Pose: %_" % [pose(netsw)])
```

Results in the following error: 

```
Uncaught Exception: Must get pad definition with a pad (received a single instance).
  in ethernet-io/network-switch
    at src/network-switch.stanza:30.24
  in ethernet-io/main
    at /Users/callendorph/designs/ethernet_io/src/main.stanza:39.7
  in ethernet-io/main
    at /Users/callendorph/designs/ethernet_io/src/main.stanza:63.0
An error occurred during execution.
```

Exception is on the attempt to call `pose(netsw)`

tjknoth:
you can use `package-poses(self)` to get the poses of any code-placed instances in some module.
But the semantics of this are a bit weird -- it won't search any module instances, you'll have to do that manually right now.

tjknoth:
And @callendorph  -- `pose` is only defined on pads for some reason. You'll have to get the package of the inst if you want to use `pose`

tjknoth:
We might want to rethink these APIs or at least write some nice abstractions... it's definitely pretty awkward as is

callendorph:
I've created a ticket - lets continue further discussion there.

callendorph:
I'm going to retract this and refactor so that I can move forward with ethernet IO without stacking multiple PRs.

PR #118, opened 2024-08-06
[Bugfix] Move courtyard origin to a custom layer.

callendorph:
Closes PROD-413

This is required because KICAD won't allow other shapes than the boundary shape on the courtyard layer.

This will allow export to complete successfully.


bhusang (approved):

callendorph:
Tests now pass locally for me.

PR #119, opened 2024-08-07
Add introspection utilities for port -> pad mappings.

callendorph:
Closes PROD-419

Adds some tools for getting the pad or pads associated with a particular port a component. This will help with extracting pad shapes for generating thermal via patterns. 

Thanks @ariel-hi  - I think this path is going to work well.

### DIFF
--- a/src/design/introspection.stanza
+++ b/src/design/introspection.stanza
@@ -134,20 +134,23 @@ This function will retrieve all of the
 @return Tuple of KeyValue mappings between `Pin` and a Tuple of `Pad` objects.
 <DOC>
 public defn port-to-pads (dev:JITXObject) -> Tuple<KeyValue<JITXObject, Tuple<JITXObject>>> :
-  ;Get the landpattern definition of the instance's component.
-  val lp = landpattern(instance-definition(dev))
+  inside pcb-module:
+    ;Get the landpattern definition of the instance's component.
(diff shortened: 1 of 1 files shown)


src/design/introspection.stanza
```diff
@@ -89,3 +90,89 @@ public defn get-element-ports (obj:JITXObject) -> [JITXObject, JITXObject] :
...
+public defn port-to-pads (dev:JITXObject) -> Tuple<KeyValue<JITXObject, Tuple<JITXObject>>> :
+  ;Get the landpattern definition of the instance's component.
+  val lp = landpattern(instance-definition(dev))
+
+  ;Build a table storing each landpattern pad by its ref.
+  val pad-table = to-hashtable<Ref, JITXObject>(seq({ref(_0) => _0}, pads(lp)))
+
+  ;Scroll through pin property table to map instance port to landpattern pads.
+  for pt in pins(dev) map :
+    val pads = for pad-ref in property(pt.pads) seq :
```
ariel-hi, line 145: To be safe, we should call `get-property?` here instead which returns a Maybe.
It'll be None for components that do not use a pin-properties table (and technically for modules too).

callendorph:
@ariel-hi  - I've updated - can you please review / approve at your convenience?

ariel-hi (approved):
Great!

PR #120, opened 2024-08-07
Refactor PCIe to use the `SI-Constraint` model

callendorph:
I've update the PCIe protocol and example to use the `SI-Constraint` model.

Here is the straight through version (ie IC to connector):
Screenshot 2024-08-07 at 9 29 26 AM

Notice the blocking caps on only the TX side. 

Here is the null-modem case (ie, IC to IC, TX => RX, RX => TX):

Screenshot 2024-08-07 at 10 57 55 AM

Blocking caps on both TX channels. Notice that the routing structure application is not working for the null-modem case because of the use of `node` in `reverse-pcie-lanes`.


bhusang (approved):
Very nice.

PR #124, opened 2024-08-07
Updated tests to add a faster test run target.

callendorph:
The current `tests` target takes 2 mins to run.
The `all-tests` added in this commit takes 6 seconds.


bhusang (approved):

callendorph:
Jason has already merged something like this.

PR #127, opened 2024-08-08
Add differential GPIO bundle

alijitx:


bhusang (commented):
What's the use case for this? And why would `diff-pair` not be usable?

callendorph:
I think we introduced this because we wanted to differentiate digital diff-pairs and analog diff-pairs.
I think it also mirrors of the structure for `gpio` single ended. 

That being said - I was mistaken when I was talking about the quasi-inheritance on bundles and Patrick set me straight. The quasi inheritance only happens when you have a parameterized bundle.

alijitx:
Should I cancel this PR then and use regular `diff-pair` instead?

alijitx:
Specifically I wanted to add this to use in the FPGA complex component generator

callendorph:
In the FPGA use case, I think we need to be able to differentiate:
1.  Digital diff-pair
2.  Analog diff-pair
3.  Clock diff-pair
4.  High-speed Transceiver pairs (and sub-families)

In general, the usage of `diff-pair` is very context dependent. Many of our current use cases are digital communication busses. I think the FPGA is the first case where we have had to differentiate between digital and other types of signals. 

I think we can assume that the digital diff-pair will be the most common in the FPGA and matches with other existing use cases. We can probably use `diff-pair` their with little trouble. We probably want separate types for analog, clock, and high speed.

bhusang:
What about adding the following then?
```
public pcb-bundle analog-diff-pair :
  port P
  port N

public pcb-bundle clk-diff-pair :
  port P
  port N

public pcb-bundle hs-diff-pair : ; not sure about this one...
  port P
  port N
```

bhusang (approved):

PR #145, opened 2024-08-26
Add Circuit Network Generator

callendorph:
Tool for constructing series, parallel, and other lumped circuit networks.

```

val circ = create-circuit( R1 + Tap(R2, `divOut)) 

```

Creates an instantiable for a resistive divider with a central tap port `divOut` and endpoint ports `p[1]` and `p[2]` like a standard two-pin component. For example:

```
pcb-module representative-circuit:
  port p : pin[[1, 2]]
  port divOut 

  inst r1 : R1 
  inst r2 : R2 
  net (r1.p[1], p[1[)
  net( r1.p[2], r2.p[1], divOut)
  net( r2.p[2], p[2] )
```

This is a simple example - but more complex circuits with `Parallel` and `ShuntTo` are possible.

### DIFF
--- a/src/circuits/Network.stanza
+++ b/src/circuits/Network.stanza
@@ -64,7 +64,9 @@ instantiated in the resultant network definition.
 <DOC>
 public defstruct CircuitElement <: ANT:
   doc: \<DOC>
-  Elements of this node - This Empty by default for `CircuitElement`
+  Sub-Elements of the `CircuitElement`
+  A CircuitElement is the Element so this collection is empty by default.
+  We construct the element from the `elem-type` object.
   <DOC>
   elements:IndexedCollection<ANT> with:
     as-method => true
--- a/src/circuits/Network.stanza
+++ b/src/circuits/Network.stanza
@@ -1,0 +1,493 @@
+#use-added-syntax(jitx)
+doc: \<DOC>
+@brief Lump Circuit Networks
+
+The concept of this code is to provide a means for defining a lumped
+circuit network tree. The idea is to provide a "DSL" for defining
+components that are used together to form a circuit network.
+
+In this graph there are Elements and Nodes.
+
+* Elements are the edges of the graph - the resistors, capacitors, etc.
+* Nodes are the pins, ground, nets, etc of the circuit
+
+We can think of the circuit network as a graph. We have the following
+ operations to this graph for 2-pin elements
+
+ 1.  Series (+) - string multiple elements in series
+ 2.  Parallel (|) - put multiple elements in parallel
+ 3.  Tap - For a particular node in the circuit graph, connect this node to
+     an external port on the resultant network instantiable (pcb-module). The
+     components are still connected on both sides in the network.
+ 4.  ShuntTo - Connect a given node to a port through an element.
+     This allows connecting a component that escapes the network. For example,
+     this might be used connect a particular component or sub-network to ground.
+ 5.  Invert (~) - Instead of connecting 1 -> 2 , connect 2 -> 1. reverses the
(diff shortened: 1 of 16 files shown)


src/circuits/Network.stanza
```diff
@@ -0,0 +1,493 @@
...
+val DEF_ELEM_PUBLIC = false
+doc: \<DOC>
+Circuit Element for a Circuit Network
+
+This type typically includes an instantiable that will be
+instantiated in the resultant network definition.
+<DOC>
+public defstruct CircuitElement <: ANT:
+  doc: \<DOC>
+  Elements of this node - This Empty by default for `CircuitElement`
```
bhusang, line 67: I am not sure that I understand this sentence.

src/circuits/Network.stanza
```diff
@@ -0,0 +1,493 @@
...
+
+```
+p[1] -> O -- a -- O -- b -- O -> p[2]
+                  |
+                  ---> divOut
+```
+
+A and B are connected in series as normal and their
+intermediate node is extracted to a port.
+
```
bhusang, line 141: To me, so YMMV, the following would be a bit more intuitive:
```
val circ = create-circuit(a + Tap(`divOut`, b))
```
notice I flipped the order of arguments in the `Tap()` function. As I read this line from left to right, I see `a` `divOut` and then `b` which is kinda the logical order of the constructed circuit. This is just a suggestion and we return you to the regularly scheduled program.

callendorph: Ya- I can see the argument. I had `ShuntTo` as well, and it made a little more sense as: 

```
val circ = create-circuit(A + (B | ShuntTo(C, `COMMON)))
```

Where `p[2]` of `C` gets connected to `COMMON` port. But we can make the `Tap` the other way. It will be more consistent with the default for `Tap` with is to make the tap port connect to `p[1]`

src/circuits/Network.stanza
callendorph: Oops

src/circuits/Network.stanza
```diff
@@ -0,0 +1,493 @@
...
+This type will not work with `Series` - it is only
+possible to use this in combination with the Parallel `|` operator.
+
+
+Example:
+This circuit:
+```
+A + (B | ShuntTo(C, "named-shunt-port"))
+```
```
bhusang, line 229: The order of arguments to `Shunt()` make sense to me.

src/circuits/Network.stanza
```diff
@@ -0,0 +1,493 @@
...
+    ; Return the first pin and the last pin
+    [pinSets[0][0], pinSets[cnt - 1][1]]
+
+doc: \<DOC>
+Instantiate and Connect all parallel elements
+
+@param a Parallel Network Element
+@param taps Accumulator for the external ports of the network.
+@return Tuple of `[p1, p2]` which is typically the first parallel elements ports. All of
+the other element's `p1` and `p2` are netted to t he first parallel element, respectively.
```
bhusang, line 341: Minor edit:
```
the other element's `p1` and `p2` are netted to t he first parallel element,
```
to 
```
the other element's `p1` and `p2` are netted to the first parallel element,
```

src/circuits/Network.stanza
```diff
@@ -0,0 +1,493 @@
...
+`ShuntTo` elements are handled differently - typically p1 of the shunted component is connected
+by the `p2` is not connected.
```
bhusang, line 343: This sentence needs some clarification.

bhusang (approved):
Cool, a reasonable syntax for circuit composition.

PR #155, opened 2024-09-05
topo-bus

tjknoth:
Implements `topo-bus` as discussed. Possible areas needing feedback/improvement:
* does not allow concrete instances as "bus segments"
* docs are not good
* returns a giant unstructured tuple of everything it instantiates. This could be more useful.


src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
     (p2:JITXObject):
       topo-net(p1, p2)
 
+doc: \<DOC>
```
callendorph, line 92: Let's move this into `src/si/bus.stanza` similar to `pairs.stanza` for `topo-pair`.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+A wrapper for `KeyValue` indicating that the given pair of
+objects should be connected by a short trace.
+<DOC>
+public deftype ShortTrace <: KeyValue
```
callendorph, line 96: Clever - I like it.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+    defmethod value (this) : value(kv) 
+defmethod print (o:OutputStream, st:ShortTrace) :
+  print(o, "ShortTrace(%_ => %_)" % [key(st) value(st)])
+
+doc: \<DOC>
+  This function will construct chained topology net consisting of
+  a sequence of "bus segments".
+
+  Given a sequence of key-value pairs of "bus segments", this function
+  will connect them, generate topology segments, and constrain the trace
```
callendorph, line 109: When you say `constrain the trace length` here - what does that mean? Are you just referring to the `short-trace` concept ? 

I guess I would like to disentangle the `delay matching` of SI from `short-trace`. May be we don't mention it here but mention it in a follow on section to talk about what `ShortTrace` does.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+  length as necessary.
+
+  A "bus segment" is either:
```
callendorph, line 112: `segment` here implies a start, end and edge of the topology. I think we might want to consider using `node` here instead of `segment`.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+  * A tuple of single-pin or array ports
+  * A single-pin or array port
+  * An Instantiable
+  * A "ShortTrace"
+
+  The left- and right-most components of the chain must be concrete --
```
callendorph, line 118: `must be concrete` - Are require ports allowed for the left-/right- most nodes ?

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+    (x:Instantiable) :
+      connect-topo-bus(lhs, x, acc, short-trace?)
+
+defn connect-topo-bus (lhs:Tuple, rhs:Instantiable, acc, short-trace?:True|False) -> Tuple :
+  inside pcb-module :
+    to-tuple $
+      for l in lhs seq :
+        inst i : rhs 
+        add(acc, i)
+        topo-net(l, i.p[1])
```
callendorph, line 182: Let's use `get-element-ports` here so that we can use both `p[1]/p[2]` and `a/c` variant two pins.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+      val [l, r] = unpack(x)
+      make-topo-bus(l, r, acc, short-trace? = x is ShortTrace)
+    (x:JITXObject) : flatten-obj(x)
+    (x:Tuple) : to-tuple $ seq-cat(flatten-obj, x)
+
+defn flatten-obj (x:JITXObject) -> Tuple<JITXObject> :
+  match(port-type(x)) :
+    (t:SinglePin) : [x]
+    (t:PortArray) : to-tuple $ pins(x)
+    (otherwise) : throw $ Exception("nope")
```
callendorph, line 208: I've been using `ValueError` here to have something consistent.

callendorph: Maybe a `TypeError` would be more reasonable here.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+
+
+; [TODO] : return val
+; [TODO] : short-trace
+defn topo-bus (lhs:JITXObject, instantiable:Instantiable, rhs:JITXObject) -> False :
```
callendorph, line 213: I assume this just gets deleted.

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+
+  @snip-note 3 `create-resistor` returns an `Instantiable`
+  @snip-note 4 `sw-bus.rxd` and `mii.rxd` are both arrays of two pins. `topo-bus` will return a tuple of four new resistors that connect the two buses.
+  @snip-note 6 All objects in `rx-bus` will be connected to their respective resistor with a short-trace.
+<DOC>
+public defn topo-bus (pair:KeyValue) -> Tuple<JITXObject> :
+  ensure-valid-tree!(pair)
+  val [l, r] = unpack(pair)
+  val acc = Vector<JITXObject>()
+  make-topo-bus(l, r, acc, short-trace? = pair is ShortTrace)
```
callendorph, line 152: So I think this means that I could do something like:

```
topo-bus(ShortTrace(A, B => C => D))
```

It seems like this would apply only to `A => B` if I'm reading the code correctly - but I'm not sure. It seems like the top level `short-trace?` will propagate through the recursion beyond the first step in multiple places of `connect` but not in the `KeyValue` case. 

Or will `ShortTrace` in this case apply to every single connection ?

src/si/helpers.stanza
```diff
@@ -88,6 +89,143 @@ public defn topo-net (pair:KeyValue<JITXObject,JITXObject|KeyValue>):
...
+  port mii : rmii-bus
+  require sw-bus : rmii-bus from mcu
+  val series-R-type = create-resistor(resistance = 51.0, precision = (1 %), case = ["0603"])
+  val rx-bus = [sw-bus.rxd, sw-bus.crs-dv, sw-bus.ref-clk]
+  val rx-out = [mii.rxd, mii.crs-dv, mii.ref-clk]
+  topo-bus(ShortTrace(rx-bus => series-R-type) => rx-out)
+  ```
+
+  @snip-note 3 `create-resistor` returns an `Instantiable`
+  @snip-note 4 `sw-bus.rxd` and `mii.rxd` are both arrays of two pins. `topo-bus` will return a tuple of four new resistors that connect the two buses.
```
callendorph, line 145: `both arrays` maybe we say `both PortArray instances of two pins`. I want to avoid confusion between `array` and `tuple` for new users.

src/si/helpers.stanza
callendorph: If we do that, we may want to have some mechanism to invert - perhaps we leave that the `Network` code and the user can just instantiate a network with a single diode and an invert:

```
  val flip-diode = create-circuit( (~ D))
  topo-bus(rx-bus => flip-diode => out-bus)
```

Or something similar. 

I guess the thing that I want to avoid is if a user puts a Diode into this `topo-bus` chain - I don't want them to get some `Can't find port p[1]` error  that they won't be able to discern.

callendorph (commented):
Looks great - Thanks @tjknoth !

I made several comments - let me know if you want to discuss anything.

src/si/helpers.stanza
tjknoth: yeah -- fixing all the exceptions today, they're all useless right now

src/si/helpers.stanza
tjknoth: yes they are. Instantiables are not. I'll change the word or clarify.

src/si/helpers.stanza
tjknoth: Yeah, I meant it to only apply to `A => B` here but I probably screwed that up.
It also seems plausible that we might want it to apply recursively to every connection?

tjknoth: One side effect of my implementation is that these two statements are equivalent:
`topo-bus(ShortTrace(A => B) => C)`
`topo-bus(ShortTrace(A => B => C))`
Seems like this might be confusing

src/si/bus.stanza
```diff
@@ -0,0 +1,174 @@
...
+  * A tuple of single-pin or array ports
+  * A single-pin or array port
+  * An instance
+  * An Instantiable
+  * A "ShortTrace"
+
+  If given an Instantiable for one node of the chain, this function will 
+  create as many concrete instances as necessary.
+
+  The left- and right-most nodes of the chain must be pins from the JITX
```
callendorph, line 51: Let's please use `ports` here instead of `pins`.

src/si/bus.stanza
```diff
@@ -0,0 +1,174 @@
...
+  design. Both concrete and abstract require pins work. They cannot be
+  instances or instantiables.
+
+  We assume that all concrete segments are the same size and shape:
+  we can only compute `topo-bus([a, b, c], [d, e, f])` when port `a`
```
callendorph, line 56:
```
topo-bus([a, b, c], [d, e, f])
```

Did you mean to use a `,` here. Or should it be a `=>`?

callendorph (approved):
Looks great. Let's make the update to the `ShortTrace` behavior we discussed and then lets move forward.

src/si/bus.stanza
tjknoth: good catch

PR #158, opened 2024-09-11
Add DQS pin option for SPI bundle

alijitx:


bhusang (commented):
So I don't think that this strobe is part of the standard SPI bus. What you may want to do instead is to define 
```
public pcb-bundle spi-with-strobe()
  ...
```
so as to not confuse the SPI bundle with additional manufacturer-specific signals. @callendorph has commented on this in the past.

bhusang (approved):

PR #159, opened 2024-09-11
Add Bypass and Pullup Tools

callendorph:
For example: 

```

  val fine-C = create-capacitor(
    bypass-query,
    capacitance = 0.1e-6,
    rated-voltage = Interval(max-rail-V, max-voltage)
  )

  for p in pins(C.VDDIO) do:
    net (p, rail-IO.V+)
    val c = insert-bypass(p, GND, elem-type = fine-C)
    schematic-group(c) = bypass-cap-IO

```

```
  insert-pullup(ROM.C.CS, VDD-3v3)
  insert-pullup(ROM.C.CLK, VDD-3v3)
  insert-pullup(ROM.C.DO, VDD-3v3)
```

This functions work with both nets, SinglePin ports, and `power` bundle ports. 

-------

I've added `maybe-utils` as a dependency. I think it could help reduce complexity in some of the code.

### DIFF
--- a/slm.toml
+++ b/slm.toml
@@ -1,3 +1,4 @@
 name = "jsl"
 version = "0.5.1"
 [dependencies]
+maybe-utils = { git = "StanzaOrg/maybe-utils", version = "0.1.6"}
--- a/src/circuits/Bypass.stanza
+++ b/src/circuits/Bypass.stanza
@@ -1,0 +1,104 @@
+#use-added-syntax(jitx)
+defpackage jsl/circuits/Bypass:
+  import core
+  import jitx
+  import jitx/commands
+
+  import jsl/bundles
+  import jsl/si/signal-ends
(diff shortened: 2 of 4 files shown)


bhusang (commented):
While I am ok with `Pullups` as the name, will we also have a `Pulldowns` as the complementary function? Or should we try to use `Straps`?

callendorph:
I don't really like `Straps`. 

I think we can include both pullups and pulldowns in `Pullups`. I think it is better than `utils` which isn't descriptive enough. I think the concept is pretty well understood. We can certainly expand the documentation to help make it more searchable.

bhusang:
Ok, then if we add `Pulldowns()` that would be ok with the requisite documentation.

bhusang (approved):
With the addition of `Pulldowns()`, it's good to go.

PR #165, opened 2024-09-12
Cja/update resistor capacitor symbols

callendorph:
This makes some improvements to the resistor and capacitor symbol generator code. 

I've also fixed a bug in the polarized capacitor rendering: 

Broken:
Screenshot 2024-09-12 155854

Now Fixed
Screenshot 2024-09-12 155909

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 SLM_STANZA ?= jstanza
 STANZA := $(SLM_STANZA)
 CWD := $(shell pwd)
+SLM := $(SLM)
 
 # execute all lines of a target in one shell
 .ONESHELL:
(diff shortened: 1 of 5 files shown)


callendorph:
@jwatson0  - I've added a new dependency to JSL's `slm.toml` for `maybe-utils`. When the tests run - I don't see a call to `slm` to retrieve the dependencies: 

https://github.com/JITx-Inc/jsl/actions/runs/10839998507/job/30081487228?pr=165#step:5:690

Is this expected ?

bhusang (approved):

jwatson0:
The JSL Makefile doesn't currenty do anything with slm.  It just calls jstanza on the tests.

Should we integrate the tests into an slm project?

`stanza.proj` also doesn't have build targets for `main` or `test`, so it seems that JSL isn't a full slm library yet.
I can add that if that's the desired step, which I assume it is.

callendorph:
> stanza.proj also doesn't have build targets for main or test, so it seems that JSL isn't a full slm library yet.
I can add that if that's the desired step, which I assume it is.

We moved away from that approach because running the static binary build takes forever. The `run-test` mechanism was much much faster (ie 5 seconds vs 30+seconds). 

That is why it is in the `Makefile` and not in the `stanza.proj` file. 

Re `slm` - I think we may need to add a `run-test` command to `slm` if it doesn't already exist.

jwatson0:
> much much faster
Oh, right, I remember now.

I also think we need a `fetch` command in slm to fetch all of the dependencies and not do any compilation.  Maybe that would help in this case.

callendorph:
>I also think we need a fetch command in slm to fetch all of the dependencies and not do any compilation. Maybe that would help in this case.

Agreed - that might be a more flexible model.

callendorph:
I have an idea for how to work around it - I need to debug it on another computer but let me see if I can get it work

callendorph:
OK - @jwatson0  - it doesn't work because the docker container does not seem to have `slm` in it. 
Is there a way that we can add `slm` to this docker container ?

callendorph:
Closed in favor of #167

PR #172, opened 2024-09-23
Phase out `get-connected-net`

tjknoth:
Eventually we will probably delete this, for now I:
* added an introspection `get-connected-nets` that returns all connected net statements
* changed `make-via-grid` to just net directly to the target object
* changed the corresponding field in `PortInfo` to a tuple
* added a function `connected-net` on `PortInfo` that mimics the old functionality (for backwards compatibility)
Another thing we will need to look at:
`make-thermal-vias` assumes the existence of a single `net` statement for its `geom` statement. This will often not be true -- it still works since I reimplemented `connected-net`, but it will arbitrarily select the first net statement. Maybe this is fine, I'm not 100% sure

### DIFF
--- a/src/design/introspection.stanza
+++ b/src/design/introspection.stanza
@@ -481,6 +481,58 @@ public defn get-connected-net (pt:JITXObject, cxt:JITXObject = self) -> Maybe<JI
         None()
 
 doc: \<DOC>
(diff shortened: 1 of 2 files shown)


callendorph (approved):
:+1: Thank you for catching this!

PR #173, opened 2024-09-23
Updates to the Silkscreen Generation

callendorph:
I'm trying to setup a framework by which we can allow customization of silkscreen content including the outline, pin1 markers, etc. 

This PR adds a interface type for silkscreen outlines and then converts existing code to create multiple different implementations of that interface.

This allows me to consolidate some code as well as set up a future where the silkscreen outline is an argument to the package. 

I've also added a better silkscreen for the SMTs that uses two lines instead of the full line rectangle:

image

It doesn't look great for small case sizes like 0402 and 0201 - we will improve those.


bhusang (approved):

callendorph:
Alright - I've fixed the small component outlines: 

image

PR #175, opened 2024-09-25
Support vias in topologies

tjknoth:
instantiate a phantom pin per-via to enable usage in topologies.
Shouldn't be merged until a jitx-client bugfix gets through.


callendorph:
I tried running the diff-pair via structure example on 3.17.0 in `examples/via-structures/diff-via-structure.stanza` and I had trouble routing:

image

The diff insertion point I can't route to the diff-pair via structure.
Also note that when I show the labels I get `p` and `p` - it would be very useful if the labels showed `p` and `n` 
When I select pads and press `a` - I see no other pads highlighted - so I think there might be an issue with the example at least or the generator at worst.

tjknoth:
I forgot some net statements -- fixed that, which I think should enable routing.
Also changed the names to `n` and `p`.


> When I select pads and press `a` - I see no other pads highlighted - so I think there might be an issue with the example at least or the generator at worst.

Do you mean when you select one of the two pads `n` or `p`? That's still happening even after i fixed the nets. Still poking.

tjknoth:
Actually, I think the possible ends thing might be an issue with the example.
If you select one of the pads on the capacitors and select possible ends (ctrl+a), it will highlight the pads on the diff pairs.
Is there a different behavior you expect?

callendorph:
Diff Pair insertion point example seems to be working in 3.19.0:

Screenshot 2024-10-17 at 10 12 14 AM

tjknoth:
Yep -- it won't work if you run from a clean state until 3.20 due to the placer bug.
I'll get this merged today anyways.

src/via-structures.stanza
```diff
@@ -549,10 +553,28 @@ defmethod make-via-structure (v:DifferentialViaStructure -- pose:Pose = DEF_VS_L
...
-    topology-segment(sig-in.P, sig-out.P)
-    net via-net-N (sig-in.N, sig-out.N)
-    topology-segment(sig-in.N, sig-out.N)
+    val [via-def-P, via-def-N] = match(via-def(v)):
+      (obj1:Via): [obj1, obj1]
+      (obj2:[Via, Via]): obj2
+    val via-pitch = pitch(v) / 2.0
+
+    ; First wrapper instance
+    inst via-pin-P : topo-via-pin(`p)
```
callendorph, line 562: The only concern I have here is the symbol for the diff-pair via structures being two separate schematic symbols in this approach. I think that will make the schematic harder to read. Assuming the user uses the `create-via-structure` function - then at least these two symbols will be in the same sch group so they will be group together a bit. 
I think this is fine for now until we can integrate via structures more directly in the JITX runtime.

callendorph (approved):
:+1: Thanks for your help on this.
The unit tests pass when I run them locally.

callendorph:
Seems like the CI is fixed now too 🎉 🦖

PR #177, opened 2024-09-26
Add Keepout Generators for SMT Components

callendorph:
This provides a mechanism for customizing the keep out generators for components, specifically SMT components: 
image

Notice that this keepout currently works for planes but it is not yet working for traces.

Goal will be to expand this to support QFNs and other packages where we need to provide keepout functionality.

### DIFF
--- a/src/ensure.stanza
+++ b/src/ensure.stanza
@@ -134,3 +134,7 @@ public defn ensure-in-set! (accepted:Tuple<Int>):
     if not contains?(accepted, value):
       throw $ ValueError("Value '%_' not in Accepted Value Set: [%,]" % [value, accepted])
   ensure-func
+
+public defn ensure-not-empty! (field:String, value:Collection) :
+  if count(value) == 0:
+    throw $ ValueError("Expected Non-Empty Collection - Found: %_" % [value])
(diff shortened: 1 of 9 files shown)


src/ensure.stanza
```diff
@@ -133,4 +133,8 @@ public defn ensure-in-set! (accepted:Tuple<Int>):
   ):
     if not contains?(accepted, value):
       throw $ ValueError("Value '%_' not in Accepted Value Set: [%,]" % [value, accepted])
-  ensure-func
\ No newline at end of file
+  ensure-func
+
+public defn ensure-not-empty! (field:String, value:Collection) :
+  if count(value) == 0:
+    throw $ ValueError("Expected Non-Empty Collection - Found: %_" % [value])
```
bhusang, line 140: What does the argument `field:String` do here?

bhusang (approved):

src/ensure.stanza
callendorph: This is the interface for the `ensure => FUNC` members of a `defstruct` field. I believe it gets used by the `defstruct` macro to tell the user which field of the defstruct had this fault case.

PR #178, opened 2024-09-29
Update BGA-simple.stanza: add example without balls missing from landpattern as syntax is somewhat unintuitive

Johnsel:
Added a comment with the syntax for a full ball grid without any inactive pads, as it took me a while (and some help fom Frederik) to figure it out myself.


examples/landpatterns/BGA-simple.stanza
```diff
@@ -45,6 +45,9 @@ pcb-component test-BGA:
   val planner = Perimeter-Matrix-Planner(
     ; Pad A2 is not present
     inactive = to-pad-island(grid, "A[2]"),
+    ; To get a full grid, set inactive to empty set
+    ; it is a required argument and can't be left out entirely:
+    ;inactive = []
```
bhusang, line 50: Yeah, the example for a full matrix should also be present using `Full-Matrix-Planner`. Thanks for the feedback.

callendorph:
@jwatson0  - The CI tests failed to run after I approved through the Github UI. Can you take a look at the logs when you get a chance ?

jwatson0:
Because this is coming from a fork outside our JITX-Inc organization, it doesn't have access to the credentials required by the build.  Can we make this into a branch in our repo instead of a fork?

callendorph:
I think @alijitx was going to submit a simple full matrix BGA landpattern

PR #180, opened 2024-10-01
Fix JSL docs

PhilippeFerreiraDeSousa:


src/layerstack.stanza
```diff
@@ -487,11 +487,11 @@ layers if any.
 @param index Conductor layer to inspect. This is a zero-based index
 where the top layer is 0, the first inner layer is 1, etc.
 <DOC>
-public defn get-conductor (ls:LayerStack, index:Int) -> [Maybe<LayerSpec>, LayerSpec, Maybe<LayerSpec>] :
+public defn get-conductor (stack:LayerStack, index:Int) -> [Maybe<LayerSpec>, LayerSpec, Maybe<LayerSpec>] :
```
callendorph, line 490: we can't use `stack` here because it is a JITX statement keyword.

callendorph (approved):
:+1: Looks good. 
There is one place to double check - the `stack` keyword I had trouble with so we might want to avoid that.

src/layerstack.stanza
PhilippeFerreiraDeSousa: The file is loading up
image

PhilippeFerreiraDeSousa: What is `stack` doing in stanza?

PhilippeFerreiraDeSousa:
```
stanza> val stack = 1
stanza> defn stack (x:Int) : println(3)
stanza> 
stanza> stack(6)
3
false
```

plasmajitx: I think the point is not that it's illegal, but rather that it's potentially confusing and dangerous.

plasmajitx: in esir:
```
  ; stack(thickness, material)
  defrule exp4 = (stack(?thickness:#exp!, ?material:#exp!)) :
    val template = `(make-stackup-layer(JITX-CLIENT, info, false, thickness, material, false))
    ir-fill $ substitute(template, [
      `info => info-exp(closest-info())
      `thickness => thickness
      `material => material])
```

PhilippeFerreiraDeSousa: Come on, you are made out of sugar xD
I don't mind whatever the name is, code updated.

PR #181, opened 2024-10-02
Pin Models for `topo-bus` instantiated components

callendorph:
This is a bit of a hack that is needed until we can address PROD-530
This allows SI topologies to properly construct constraints when using topo-bus.


callendorph:
I have to move forward to meet deadlines so I'm going to merge. Let me know if there is anything we need to discuss and I can fix it in a follow up PR.

PR #182, opened 2024-10-02
Added more functions for introspecting pads of a component

callendorph:
These tools are for constructing keepouts around components or parts of components, typically from a wrapping module.

### DIFF
--- a/src/design/introspection.stanza
+++ b/src/design/introspection.stanza
@@ -5,6 +5,8 @@ defpackage jsl/design/introspection:
   import jitx
   import jitx/commands
 
+  import maybe-utils
+
   import jsl/errors
   import jsl/landpatterns/introspection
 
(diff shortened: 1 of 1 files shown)


callendorph:
I need to more forward to meet deadlines. I'm going to merge. Please let me know if there is anything I need to address in a follow on PR before a release.

src/design/introspection.stanza
```diff
@@ -269,6 +271,21 @@ public defn get-pads-from-port (pt:JITXObject, cxt:JITXObject = self) -> Maybe<T
...
+  for pt in pts seq-cat:
+    val pds? = get-pads-from-port(pt, cxt)
+    value-or(pds?, [])
+
+doc: \<DOC>
+Convert a Sequence of JITXObjects to LandPatternPad objects.
+<DOC>
+public defn to-lp-pad-seq (objs:Seq<JITXObject>) -> Seq<LandPatternPad> :
+  for obj in objs seq:
+    obj as LandPatternPad
```
tjknoth, line 287: Only possible change is to match and throw an exception instead of cast -- slightly scary to expose a blind cast via public API

callendorph: OK - perhaps I'm being too cavalier with the `as` operator. Will that `fatal` if it doesn't match correct ?

callendorph: Ya - so I think this function should be in JITX not JSL if anywhere.
In this particular case, I'm confident that this JITXObject is a LandPatternPad. I could put this code in the design but it is going to be duplicated everywhere I use this pattern. 
The other option is to change the `bounds` function to accept a `JITXObject` but that really just moves the problem. When `bounds` tries to inspect a LandPatternPad and it doesn't find one - then an error will be raised there. I think the `jsl/landpatterns/leads/bounds` function was probably written when I was trying for typing - but perhaps I need to re-evaluate that instead of adding this function.

PR #188, opened 2024-10-04
add enum variant docs to jsl

plasmajitx:


bhusang:
@plasmajitx This is great, but how does one know that all @members of the various objects in JSL are listed? Do you have an automated method for discovery?

bhusang (approved):

plasmajitx:
docgen prints a warning when it thinks there's an enum variant that doesn't have an `@member`.  so I had a big list of warnings to work from.

it also relies on having an `@member` internally to know that the definition is a true defenum and to resolve links to the variants, which is why I wanted to make sure we get all these in.

PR #189, opened 2024-10-09
fix ambiguous links

plasmajitx:
- **fix ambiguous type/func links in jsl docstrings**
- **fix disambiguated links that weren't actually types**


ariel-hi (approved):
⭐

PR #193, opened 2024-10-15
Keepout Box Bugfix

callendorph:
A query error was selecting for super small components and resulted in exposing an error in the keepout logic - especially for components where the landpattern soldermask openings overlap. I've updated the logic to better handle that case as well as change the keepout interior pad boundary to use the copper shape instead of the soldermask.

### DIFF
--- a/src/landpatterns/VirtualLP.stanza
+++ b/src/landpatterns/VirtualLP.stanza
@@ -888,10 +888,22 @@ public defn pad-interior-bounds (vp:VirtualLP, side:Side -- layer-spec:LayerSpec
     val row0 = get-bounds $ get-pads-by-row(vp, 0)
     val row1 = get-bounds $ get-pads-by-row(vp, 1)
 
-    Box(
-      Point(left(row1), up(row1))
-      Point(right(row0), down(row0))
-    )
+    try:
+      Box(
+        Point(left(row1), up(row1))
+        Point(right(row0), down(row0))
+      )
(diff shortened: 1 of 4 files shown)


src/landpatterns/VirtualLP.stanza
```diff
@@ -868,24 +868,42 @@ which was to try and find lines and intersections of those lines.
...
-    Box(
-      Point(left(row1), up(row1))
-      Point(right(row0), down(row0))
-    )
+    try:
+      Box(
+        Point(left(row1), up(row1))
+        Point(right(row0), down(row0))
+      )
+    catch (e:NegativeBoxError):
```
tjknoth, line 896: why isn't the Box guaranteed to be valid? does `get-bounds` not enforce any order?

callendorph: Because if you use the soldermask openings instead of the copper geometry - the soldermask openings might overlap. This isn't exactly what we want but we have to design with the specifications of the fab house. 

If the soldermask openings overlap, then this algorithm will result in creating a negative box. I don't think we want to force  create a box (ie the absolute value of the box). This is more pressing for the silkscreen case - where a negative box would indicate that we can't put silkscreen here anyway (no soldermask is present). 

For the copper case - this would technically just be an error - because it would mean that there are two pads with overlapping copper.  In that case - we would just fix the copper pads.

PR #194, opened 2024-10-15
Added a new tool for setting the config strap BOM status

callendorph:
This is a function for setting the BOM status of the components in the ConfigStrap object.


bhusang (approved):

callendorph:
Tests pass when I run them locally.

PR #209, opened 2024-11-08
Updates to the non-plated hole and support for changes to `mechanical/MountingHoles`

callendorph:
Many changes here - most of them are going to support better handling of mounting holes.

### DIFF
--- a/src/landpatterns/numbering.stanza
+++ b/src/landpatterns/numbering.stanza
@@ -45,8 +45,8 @@ doc: \<DOC>
 Null Numbering Object
 
 This is a default object for creating packages that don't require
-any pad numbering. These landpatterns may not contain any pads to
-number.
+any pad numbering. These landpatterns will likely not have any
+pads to number.
 <DOC>
 public defstruct NullNumbering <: Numbering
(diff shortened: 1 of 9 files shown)


bhusang (approved):

src/landpatterns/numbering.stanza
```diff
@@ -41,6 +41,17 @@ Convert the row and column indices into a pad number or reference
 <DOC>
 public defmulti to-pad-id (x:Numbering, row:Int, column:Int) -> Int|Ref
 
+doc: \<DOC>
+Null Numbering Object
+
+This is a default object for creating packages that don't require
+any pad numbering. These landpatterns may not contain any pads to
```
bhusang, line 48: Is this `may not` or `should not`?

callendorph: This doesn't prevent composition of landpatterns or manually defining, for example, a mechanical hole pad outside of these constructs.

bhusang: Cool, then maybe we change the language to something like? 
"These landpatterns do not need to contain any pads to number."

PR #212, opened 2024-11-14
fixed BD defn

bhusang:


callendorph (approved):
:+1: Design runs. Looks good.

PR #214, opened 2024-11-20
Set signal boundary even if the shrink parameter isn't passed

d-haldane:
PD crashes when the signal-boundary isn't set, so let's remove a footgun.


callendorph (approved):
:+1:

PR #223, opened 2024-12-10
Misc updates and Docs Fix

callendorph:
Addresses issue that was causing the documentation build to fail. 
Other minor updates.


jwatson0 (approved):
:+1:

PR #224, opened 2024-12-11
Add Docgen to CI Build

callendorph:
This will hopefully check for docstring bugs so that we can make sure those get fixed before merging.


callendorph:
docker on my OS-X brew installation is basically non-functional. I can't test locally so there may be a long string of commits here to get this working

callendorph:
Tests passed but it didn't actually run what I told it to run. Something is wrong with the CI.

callendorph:
When I run in the VM I see this:

```
...
ols/transformer jsl/symbols/transistors/BJT jsl/symbols/transistors/FET jsl/symbols/transistors/JFET jsl/symbols/transistors/generators jsl/symbols/utils jsl/examples/protocols/displayport/displayport-src jsl/examples/landpatterns/app-specific-keepout jsl/examples/protocols/common/example-board jsl/examples/protocols/sata/SATA-src jsl/examples/protocols/ethernet/10GBASE-KR/e10GBASE-KR-src jsl/examples/pin-assignment/swizzle -standalone src -o ./docs_build/docs
jitx-client-1  | Error: /app/jsl/src/si/pairs.stanza:144.12 Function 'topo-pair' has documentation for non-existent argument 'pairs'
jitx-client-1  | Error: Encountered 1 errors while reading documentation for package jsl/si/pairs
jitx-client-1  | Halting due to errors while reading package docs.
jitx-client-1  | make: *** [Makefile:92: test-docs] Error 255
jitx-client-1 exited with code 2
ubuntu@ip-172-31-18-60:~/jsl2/.github$ 
```

Which is what I want to happen - it fails on a documentation problem.

callendorph:
I'm going to create a new clean PR

PR #235, opened 2025-01-23
Update to `dp-coupler` for non-JSL 2-pin landpatterns

callendorph:
In the previous implementation 

```
  inst tx-coupler : dp-coupler(
    create-part(
      mpn = "ESD132-B1-W0201 E6327", 
      manufacturer = "Infineon"
    )
  )
```

Results in something like this:

Screenshot 2025-01-23 at 3 16 32 PM

because the parts-DB landpattern for this part is rotated 90 degrees from the JSL 2-pin standard landpattern.

This PR adds a new `pre-pose` argument that allows for the components in the `dp-coupler` to be oriented before being constructed in the coupler form. 

```
  inst tx-coupler : dp-coupler(
    create-part(
      mpn = "ESD132-B1-W0201 E6327", 
      manufacturer = "Infineon"
    ),
    pre-pose = loc(0.0, 0.0, 90.0)
  )
```

Screenshot 2025-01-23 at 3 18 47 PM

@Johnsel - FYI

### DIFF
--- a/src/landpatterns/courtyard.stanza
+++ b/src/landpatterns/courtyard.stanza
@@ -162,6 +162,24 @@ public defn get-courtyard-boundary (
   else:
     bounds $ Union(content)
 
+doc: \<DOC>
+Extract the courtyard shape for the landpattern of a component.
+
+This function introspects the given component and finds the courtyard
+layer of its landpattern. It then returns a Union of the landpatterns
+courtyard layer.
+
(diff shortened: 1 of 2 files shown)


Johnsel:
Thank you for accomodating a fix for this issue so quickly! :)

bhusang (approved):

PR #243, opened 2025-04-02
Bg/update ddr3

bhusang:


src/protocols/memory/ddr3.stanza
```diff
@@ -192,25 +191,8 @@ with:
 
 doc: \<DOC>
 Constructor for the DDR3 Data Link Constraint
-@param width Width of the DDR3 interface (enumerated type)
-@param diff-dqs-rs Differential Routing Structure constraints for DQS signals
-@param se-dq-rs Routing Structure constraints for DQ signals
 <DOC>
-public defn DDR3-Data-ACC-Constraint (
```
callendorph, line 199: Ya, removing these is a good idea, the `keyword-constructor=true` is a better approach to rely on.

callendorph (approved):
:+1: Thanks!

